PR #22, opened 2017-11-23
Merge debug-appveyor to master

j-tr:
Appveyor is now working - maybe squash merge due to lots of "test"-commits


appveyor.yml
```diff
@@ -2,9 +2,10 @@ version: '{build}'
 branches:
   only:
   - master
+  - debug-appveyor
```
Felerius: Should be removed before we merge this

appveyor.yml
```diff
@@ -2,9 +2,10 @@ version: '{build}'
...
 clone_folder: c:\projects\llassetgen
 image:
-- Visual Studio 2013
+#- Visual Studio 2013
```
Felerius: Why can't we use VS 2013 anymore? And if we can't, I think this should be documented somewhere, maybe in the Readme

appveyor.yml
```diff
@@ -21,24 +22,54 @@ matrix:
 
 # skip unsupported combinations
 init:
-- set arch=
-- if "%arch%"=="Win64" ( set arch= Win64)
```
Felerius, line 25: Why don't we need `%arch%` anymore. It gets used a few lines down

appveyor.yml
```diff
@@ -21,24 +22,54 @@ matrix:
...
-- if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2013" ( set generator="Visual Studio 12 2013%arch%" )
+- if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2017" ( 
+    set generator="Visual Studio 15 2017 %arch%"&&
+    set tools=msvc15)
+- if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2015" ( 
+    set generator="Visual Studio 14 2015 %arch%"&& 
+    set tools=msvc14)
+- if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2013" ( 
+    set generator="Visual Studio 12 2013 %arch%"&& 
+    set tools=msvc12)
```
Felerius: If we don't support VS 2013 we should remove these lines as well

cmake/FindFreetype.cmake
```diff
@@ -70,6 +70,7 @@ set(FREETYPE_FIND_ARGS
     ENV GTKMM_BASEPATH
     [HKEY_CURRENT_USER\\SOFTWARE\\gtkmm\\2.4;Path]
     [HKEY_LOCAL_MACHINE\\SOFTWARE\\gtkmm\\2.4;Path]
+	$ENV{FREETYPE_DIR}
```
Felerius: `FREETYPE_DIR` is already a path hint a few lines above, why do we need to duplicate it? According to the CMake manual, hints and paths are the same, hints are just searched before paths. Also tabs vs. spaces.

Felerius (changes requested):
I left a few comments, but looks pretty good overall. And yes, we should definitely squash this (having a commit named "first draft of Bitmap class" in this branch is somewhat confusing). Also, there was some trailing whitespace which I went ahead and removed

appveyor.yml
j-tr: arch is set by appveyor externally. line 25 only added a whitespace in front of the string which is somehow useless

appveyor.yml
j-tr: i couldnt figure out, why only the vs 2013 debug build fails, but since we have 2017 it should be ok to just drop vs 2013. I dont think this should be documented somewhere, also because it wasnt documented before that we are trying to build for vs 2013.

cmake/FindFreetype.cmake
j-tr: indeed. not necessary.

Felerius (approved):

PR #23, opened 2017-11-28
add badgets to README

anne-gropler:
https://shields.io/ is a simple way to use badges with clean API

Daniel proposed to use shields.io.
What do you think?


Blaidd-Drwg (approved):
Looks fancy :+1:

PR #24, opened 2017-11-29
Integrate clang-format

Felerius:
Adds CMake targets to run clang-format and a `.clang-tidy` file to control the formatting. The CMake code is adapted from the health checks code, there is `format-${target}` target for each library/executable, and a `format-all` target for convenience.

The settings are open for discussion, I simply copied from a previous project of mine. [See here](https://github.com/hpicgs/openll-asset-generator/compare/integrate-clang-format...clang-format-example) for what the current code formatted with the current settings would look like. The current settings are based on the google style guide, run `clang-format -style=file -dump-config` to see the full list of configured settings.

Note to self: Delete `clang-format-example` branch when this gets merged.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,5 @@
+# vi: ft=yaml
+BasedOnStyle: Google
+IndentWidth: 4
+NamespaceIndentation: All
+Standard: Cpp11
--- a/.clang-format
+++ b/.clang-format
@@ -3,4 +3,5 @@ BasedOnStyle: Google
 IndentWidth: 4
 ColumnLimit: 120
 NamespaceIndentation: All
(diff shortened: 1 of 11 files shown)


anne-gropler:
format runs fine in visual studio, it builds without any further output. If this is intended, then it seems fine. regarding the style rules: I am okay with that.

anne-gropler (approved):

Felerius:
I configured clang-format to apply the format directly to the source files. The other alternative would be to have it show a diff as the command output.

One easy way to see what clang-format changed is to stage all files with `git add`, run the formatter and then view the changes using `git diff`. I usually just do this just before I commit changes, because I need to `git add` all files anyway.

PR #26, opened 2017-12-11
Add common 2d vector type

Felerius:
Adds a common vector type for the project. I implemented deliberately few operations (plus, minus and equality comparison), should we need support for something like `std::hash` or `*` we can add it later.

### DIFF
--- a/source/llassetgen/include/llassetgen/Vec2.h
+++ b/source/llassetgen/include/llassetgen/Vec2.h
@@ -1,0 +1,63 @@
+#pragma once
+
+namespace llassetgen {
+    template <class T>
+    struct Vec2 {
+        T x = 0;
+        T y = 0;
+
+        Vec2() = default;
+
+        template <class T2>
+        Vec2(T2 _x, T2 _y) : x{_x}, y{_y} {}
+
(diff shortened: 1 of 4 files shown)


source/llassetgen/include/llassetgen/Vec2.h
```diff
@@ -0,0 +1,63 @@
...
+
+namespace llassetgen {
+    template <class T>
+    struct Vec2 {
+        T x = 0;
+        T y = 0;
+
+        Vec2() = default;
+
+        template <class T2>
```
Blaidd-Drwg: What is the second template for?

Blaidd-Drwg (approved):
LTGM :+1:

source/llassetgen/include/llassetgen/Vec2.h
Felerius: I mostly kept it because it was defined that way in the original `PositionType`. The only place I know where this is necessary is when moving the arguments (`T2&&`), but that is definitely not needed if we only plan on using integers and floats with this class.

I'll be removing the second template, we can always add it back if necessary.

PR #29, opened 2017-12-13
Feature: Distance Transform

Lichtso:


source/llassetgen/include/llassetgen/DistanceTransform.h
```diff
@@ -63,6 +63,8 @@ namespace llassetgen {
         std::unique_ptr<Parabola[]> parabolas;
         std::unique_ptr<OutputType[]> lineBuffer;
 
+        template<bool fill>
+        void edgeDetection(DimensionType offset, DimensionType pitch, DimensionType length);
```
scheibel, line 67: Usually the source code of a template function is provided with the header (we use `.inl` files that are included at the end of the header file). Since this is a private function, this will probably be no problem.

source/llassetgen/source/DistanceTransform.cpp
```diff
@@ -219,6 +219,41 @@ namespace llassetgen {
...
 
+
+
+    template<bool fill>
+    void ParabolaEnvelope::edgeDetection(DimensionType offset, DimensionType pitch, DimensionType length) {
+        InputType prev = 0;
+        DimensionType begin = 0;
+        for(DimensionType j = 0; j < length; ++j) {
+            InputType next = inputAt(offset+j*pitch);
+            if(next == prev)
```
scheibel, line 232: When is this possible?

source/llassetgen/source/DistanceTransform.cpp
```diff
@@ -219,6 +219,41 @@ namespace llassetgen {
...
+                continue;
+            DimensionType end = (next) ? j : j-1;
+            if(fill)
+                for(DimensionType i = begin; i < end; ++i) {
+                    if(begin == 0 && !prev)
```
scheibel: I don't understand why the check for `prev` is relevant.

source/llassetgen/source/DistanceTransform.cpp
```diff
@@ -219,6 +219,41 @@ namespace llassetgen {
...
+                        outputAt(offset+i*pitch) = square((i < (end+begin)/2) ? i-begin+1 : end-i);
+                }
+            prev = next;
+            begin = end+1;
+            outputAt(offset+end*pitch) = 0;
+        }
+        if(fill)
+            for(DimensionType i = begin; i < length; ++i) {
+                if(begin == 0 && !prev)
+                    outputAt(offset+i*pitch) = 1E20F;
```
scheibel: Is this some kind of infinity? `std::numeric_limits<float>::infinity()`

scheibel: Either way, it is a magic number and should be referenced through an own constant or symbol.

source/llassetgen/source/DistanceTransform.cpp
Lichtso: The loop ignores a pixel, when it has the same value (boolean) as the previous one. (Hence the name edge detection)

source/llassetgen/source/DistanceTransform.cpp
Lichtso: std::numeric_limits<OutputType>::infinity() does not work here, but I'll use std::numeric_limits<OutputType>::max() instead

PR #31, opened 2017-12-19
Fix META_CMAKE_INIT_SHA

Felerius:
It was set to the commit applying cmake-init in this repository,
but is supposed to be set to the commit in the cmake-init
repository that was applied to this one.

With this change, `make check-template` now generates a valid
diff link.

@Lichtso Could you quickly verify that the commit sha matches the one you applied?


Lichtso:
Yes, I copied it between
https://github.com/cginternals/cmake-init/commit/83d7cbc29a6fcb74a98498e5b0fcebd953d9d5cc
and
https://github.com/cginternals/cmake-init/commit/f23426d5783ca942ba71e45e6e936985b5db4a87
which ironically is exactly the same fix.

anne-gropler:
@Lichtso So could you officially 'review' this? And then merge it?

Lichtso (commented):
LGTM

Lichtso (approved):

PR #32, opened 2018-01-01
Image class functionality to be reviewed

j-tr:


source/llassetgen-cmd/main.cpp
```diff
@@ -5,14 +5,68 @@
...
 
-
-
 using namespace llassetgen;
 
 int main(int argc, char** argv) {
     llassetgen::init();
 
-    std::unique_ptr<DistanceTransform> dt(new DeadReckoning());
+	// creating the face
```
Lichtso: Please move this to:
source/tests/llassetgen-tests/main.cpp

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,293 @@
...
+		height = _max_y - _min_y;
+		stride = _stride;
+		data = _data;
+		min_x = _min_x;
+		min_y = _min_y;
+		max_x = _max_x;
+		max_y = _max_y;
+		bit_depth = _bit_depth;
+	}
+
```
Lichtso: Image::at and void Image::put are overcomplicated and inefficient (e.g. a / c + b / c = (a+b) / c)
also the idea of stride is not to be a alias for width but to include bitDepth.
Furthermore, this implementation contains a loop and is not endianness aware.
And they might be renamed to get / set instead of at / put.

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,293 @@
...
+			}
+		}
+	}
+
+	Image::Image(const size_t _width, const size_t _height, const size_t _bit_depth) {
+		assert(_bit_depth == 1 || _bit_depth == 2 || _bit_depth == 4 || _bit_depth == 8 || _bit_depth == 16 || _bit_depth == 24 || _bit_depth == 32);
+		width = _width;
+		height = _height;
+		bit_depth = _bit_depth;
+		stride = _width;
```
Lichtso: stride = _width * (_bitDepth / 8);

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,293 @@
...
+				data.get()[offset_y * stride * bit_depth / 8 + offset_x * bit_depth / 8 + byte_pos] = (uint8_t)in_int;
+				in_int >>= 8;
+			}
+		}
+	}
+
+	void Image::exportPng(const std::string filepath) {
+		std::ofstream out_file(filepath, std::ofstream::out | std::ofstream::binary);
+		if (!out_file.good()) {
+			std::cerr << "couldnt open file";
```
Lichtso: Couldn't / Could not
and printing the filepath would be nice as well

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,293 @@
...
+		for (size_t i = 0; i < height; i++) {
+			row_ptrs.get()[i] = (png_bytep)multi_channel_data.get() + i * stride;
+		}
+
+		png_read_image(png, row_ptrs.get());
+
+		png_destroy_read_struct(&png, &info, (png_infopp)0);
+		in_file.close();
+
+		data = std::shared_ptr<uint8_t>(new uint8_t[width * height * bit_depth / 8]);
```
Lichtso: stride * height

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,293 @@
...
+			PNG_COLOR_TYPE_GRAY_ALPHA
+			PNG_COLOR_MASK_PALETTE
+			PNG_COLOR_MASK_COLOR
+			PNG_COLOR_MASK_ALPHA*/
+		}
+		
+		png_read_update_info(png, info);
+		uint8_t channels = png_get_channels(png, info);
+		
+		uint8_t stride = width * bit_depth * channels / 8;
```
Lichtso: stride in this local scope here shadows the member / instance variable

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,293 @@
...
+
+		png_read_image(png, row_ptrs.get());
+
+		png_destroy_read_struct(&png, &info, (png_infopp)0);
+		in_file.close();
+
+		data = std::shared_ptr<uint8_t>(new uint8_t[width * height * bit_depth / 8]);
+		for (size_t y = 0; y < height; y++) {
+			for (size_t x = 0; x < width; x++) {
+					data.get()[y * width * bit_depth / 8 + x * bit_depth / 8] = multi_channel_data.get()[y * stride + x * bit_depth / 8 * channels];
```
Lichtso: Why not use Image::put here?

source/llassetgen-cmd/main.cpp
```diff
@@ -5,14 +5,68 @@
...
+	FT_Face face;
+	std::string font_file = "C:/Windows/Fonts/Arial.ttf"; // path to local font
+	char letter = 'A';
+	assert(FT_New_Face(freetype, font_file.c_str(), 0, &face) == 0);
```
scheibel: Beware that this code isn't executed for release executables!
I presume you want to execute this code and check the return value as assertion when in debug mode.

source/llassetgen-cmd/main.cpp
```diff
@@ -5,14 +5,68 @@
...
+	ft_8bit.exportPng("D:/Desktop/A_glyph1_out.png");
+
+	// create blank image with 40x30 pixels and 32 bit
+	Image blank_32bit(40, 30, 32);
+	// set pixels
+	blank_32bit.put(3, 4, 34567);
+	blank_32bit.put(4, 5, 1234);
+
+	// get pixels and test
+	for (int i = 0; i < blank_32bit.get_height(); i++) {
```
scheibel: Rather cumbersome code to clear the whole image and set two specific pixels.

source/llassetgen/include/llassetgen/Image.h
```diff
@@ -0,0 +1,39 @@
...
+			LLASSETGEN_API Image(const size_t width, const size_t height, const size_t bit_depth);
+			LLASSETGEN_API void exportPng(const std::string filepath);
+			LLASSETGEN_API uint32_t at(const size_t x, const size_t y);
+			LLASSETGEN_API void put(const size_t x, const size_t y, const uint32_t data);
+			LLASSETGEN_API size_t get_width();
+			LLASSETGEN_API size_t get_height();
+			LLASSETGEN_API size_t get_bit_depth();
+			LLASSETGEN_API Image view(const size_t _min_x, const size_t _max_x, const size_t _min_y, const size_t _max_y);
+		private:
+			Image::Image(const size_t _min_x, const size_t _max_x, const size_t _min_y, const size_t _max_y, const uint8_t _bit_depth, const size_t _stride, const std::shared_ptr<uint8_t> _data);
```
scheibel: Unnecessary `const` in declaration (however, for the definition the `const` is fine).

source/llassetgen/include/llassetgen/Image.h
```diff
@@ -0,0 +1,39 @@
...
+#include <png.h>
+
+#include <ft2build.h>
+#include FT_IMAGE_H
+
+namespace llassetgen {
+	class Image {
+		public:
+			LLASSETGEN_API Image(const FT_Bitmap &ft_bitmap);
+			LLASSETGEN_API Image(const std::string filepath);
```
scheibel: Parameter passing by copy? Why not by reference?

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,294 @@
...
+	uint32_t Image::at(const size_t x, const size_t y) {
+ 		size_t offset_x = x + min_x;
+		size_t offset_y = y + min_y;
+		assert(offset_x < max_x && offset_x >= min_x && offset_y < max_y && offset_y >= min_y);
+		if (bit_depth <= 8) {
+			uint8_t byte = data.get()[offset_y * stride + offset_x * bit_depth / 8];
+			size_t bit_pos = offset_x % (8 / bit_depth);
+			byte <<= bit_pos * bit_depth;
+			byte >>= 8 - bit_depth;
+			return (uint32_t)byte;
```
scheibel: There is a C++ `static_cast<std::uint32_t>()` for that.

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,294 @@
...
+			in_byte <<= 8 - bit_depth;
+
+			mask >>= 8 - bit_depth;
+			in_byte >>= 8 - bit_depth;
+
+			size_t bit_pos = offset_x % (8 / bit_depth);
+			in_byte <<= 8 - bit_pos * bit_depth - bit_depth;
+			mask <<= 8 - bit_pos * bit_depth - bit_depth;
+
+			data.get()[offset_y * stride + offset_x * bit_depth / 8] = data.get()[offset_y * stride + offset_x * bit_depth / 8] & ~mask | in_byte;
```
scheibel: Unnecessary recompution of index (`offset_y * stride + offset_x * bit_depth / 8`).

source/llassetgen/source/Image.cpp
```diff
@@ -184,40 +195,63 @@ namespace llassetgen {
...
-		for (int y = 0; y < height; y++) {
-			png_write_row(png, reinterpret_cast<png_bytep>(&data.get()[y * stride]));
+
+		if (bit_depth >= 24) {
+			std::unique_ptr<uint16_t[]> row(new uint16_t[width]);
+			// possible 32 float or 32 or 24 bit int data
+			// scale down to 16 bit int grayscale
+			
+			if (max > pow(2, bit_depth)) {
+				max = pow(2, bit_depth);
```
Lichtso: pow(2, bit_depth) is using floating point arithmetics
Use bit shifting instead: 1 << bit_depth

Lichtso:
Could you move your tests in main.cpp to the designated test file?
I can then approve your PR

source/llassetgen/include/llassetgen/DistanceTransform.h
```diff
@@ -1,46 +1,24 @@
...
-        void inputAt(PositionType pos, bool bit);
-        OutputType& outputAt(DimensionType offset);
-        OutputType& outputAt(PositionType pos);
-        OutputType outputAtClamped(PositionType pos);
-
-        LLASSETGEN_API void importFreeTypeBitmap(FT_Bitmap_* bitmap, DimensionType padding);
-        LLASSETGEN_API void importPng(std::string path);
-        LLASSETGEN_API void exportPng(std::string path, OutputType blackDistance, OutputType whiteDistance,
-                                      DimensionType bitDepth);
+        const Image &input, &output;
```
scheibel: Please split bulk declaration into separate lines.

source/llassetgen/include/llassetgen/Image.h
```diff
@@ -0,0 +1,40 @@
...
+#include <string>
+#include <memory>
+
+struct FT_Bitmap_;
+struct png_struct_def;
+
+namespace llassetgen {
+	class Image {
+		public:
+			LLASSETGEN_API Image(const std::string &filepath);
```
scheibel: Please use 4 spaces instead of tabs.

source/llassetgen/include/llassetgen/Image.h
```diff
@@ -0,0 +1,40 @@
...
+#include <llassetgen/Vec2.h>
+
+#include <limits>
+#include <string>
+#include <memory>
+
+struct FT_Bitmap_;
+struct png_struct_def;
+
+namespace llassetgen {
```
scheibel, line 13: We don't indent within a namespace and place the opening brace on the next line.

source/llassetgen/source/DistanceTransform.cpp
```diff
@@ -1,300 +1,162 @@
...
+        return posBuffer[pos.y * input.get_width() + pos.x];
     }
 
     void DeadReckoning::transformAt(PositionType pos, PositionType target, OutputType distance) {
         target += pos;
-        if (outputAtClamped(target) + distance < outputAt(pos)) {
+        if (output.is_valid(target) && getPixel<OutputType>(target) + distance < getPixel<OutputType>(pos)) {
             posAt(pos) = target = posAt(target);
-            outputAt(pos) = std::sqrt(square(pos.x - target.x) + square(pos.y - target.y));
+            setPixel<OutputType>(pos, std::sqrt(square(pos.x - target.x) + square(pos.y - target.y)));
```
scheibel: I think glm [1] would be a valid dependency for this project.

[1] https://github.com/g-truc/glm

source/llassetgen/source/DistanceTransform.cpp
```diff
@@ -1,300 +1,162 @@
...
-            for(DimensionType i = begin; i < length; ++i) {
-                if(begin == 0)
-                    outputAt(offset+i*pitch) = std::numeric_limits<OutputType>::max();
-                else if(!prev)
-                    outputAt(offset+i*pitch) = square(i-begin+1);
-                else
-                    outputAt(offset+i*pitch) = square((i < (length+begin)/2) ? i-begin+1 : length-i);
-            }
+            for(DimensionType i = begin; i < length; ++i)
+                setPixel<OutputType, flipped>({i, offset}, (begin == 0)
```
scheibel: This could profit from a comment.

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+	Image::Image(const size_t width, const size_t height, const size_t _bit_depth) {
+		assert(_bit_depth == 1 || _bit_depth == 2 || _bit_depth == 4 || _bit_depth == 8 || _bit_depth == 16 || _bit_depth == 24 || _bit_depth == 32);
+		min.x = 0;
+		min.y = 0;
+		max.x = width;
+		max.y = height;
+        bit_depth = _bit_depth;
+        stride = (size_t)std::ceil(float(get_width() * (float(bit_depth) / 8.0f)));
+		data = std::shared_ptr<uint8_t>(new uint8_t[get_height() * stride]());
```
scheibel: I suggest using `std::make_shared`.

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,358 @@
...
+		if (bit_depth <= 8) {
+			uint8_t byte = data.get()[offset.y * stride + offset.x * bit_depth / 8];
+			size_t bit_pos = offset.x % (8 / bit_depth);
+			byte <<= bit_pos * bit_depth;
+			byte >>= 8 - bit_depth;
+			uint32_t casted = static_cast<uint32_t>(byte);
+			return *reinterpret_cast<pixelType*>(&casted);
+		} else {
+			uint32_t return_data = 0;
+			for (size_t byte_pos = 0; byte_pos < bit_depth / 8; ++byte_pos) {
```
scheibel: This code looks optimized, but I presume it misses some optimization potential.

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+	Image::Image(const Vec2<size_t> _min, const Vec2<size_t> _max, const size_t _stride, const uint8_t _bit_depth, const std::shared_ptr<uint8_t>_data) {
+		min = _min;
+		max = _max;
+        stride = _stride;
+		bit_depth = _bit_depth;
+        data = _data;
+	}
+
+    bool Image::is_valid(const Vec2<size_t> pos) const {
```
scheibel: Take a const reference as parameter?

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,358 @@
...
+			abort();
+		}
+
+		png_set_write_fn(png, (png_voidp)&out_file, write_data, flush_data);
+
+		png_set_IHDR(png,
+			info,
+			get_width(),
+			get_height(),
+			(bit_depth < 16) ? bit_depth : 16,
```
scheibel: This looks like a `std::min`.

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,358 @@
...
+	}
+
+	template LLASSETGEN_API void Image::exportPng<uint32_t>(const std::string &filepath, uint32_t min, uint32_t max);
+	template LLASSETGEN_API void Image::exportPng<uint16_t>(const std::string &filepath, uint16_t min, uint16_t max);
+	template LLASSETGEN_API void Image::exportPng<uint8_t>(const std::string &filepath, uint8_t min, uint8_t max);
+	template LLASSETGEN_API void Image::exportPng<float>(const std::string &filepath, float min, float max);
+
+	void Image::read_data(png_structp png, png_bytep data, png_size_t length) {
+		png_voidp a = png_get_io_ptr(png);
+		((std::istream*)a)->read((char*)data, length);
```
scheibel: I suggest to add an assertion that a `dynamic_cast` to `std::istream*` results in a valid pointer.

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,358 @@
...
+			PNG_COLOR_MASK_ALPHA*/
+		}
+
+		png_read_update_info(png, info);
+		uint8_t channels = png_get_channels(png, info);
+
+		stride = get_width() * (bit_depth / 8);
+
+		std::unique_ptr<png_bytep> row_ptrs(new png_bytep[get_height()]);
+		std::unique_ptr<uint8_t> multi_channel_data(new uint8_t[get_height() * stride * channels]);
```
scheibel: What is the advantage of this code over a `std::vector<uint8_t>`?

source/llassetgen/source/DistanceTransform.cpp
Lichtso: Done

source/llassetgen/source/DistanceTransform.cpp
Lichtso: I tried using alternatives like `std::hypot(x, y)` but it is not that easy because of the signed / unsigned types

anne-gropler: In llassetgen-rendering I use glm, too. So if we need glm as a main dependency, I would benefit from it, too.

@Lichtso I think @scheibel meant to use glm instead of implementing vector maths by ourselfs.

source/llassetgen/include/llassetgen/Image.h
anne-gropler: In this project we use a formatting tool which places the braces like that.

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,358 @@
...
+		min.x = 0;
+		min.y = 0;
+		max.x = width;
+		max.y = height;
+        bit_depth = _bit_depth;
+        stride = (size_t)std::ceil(float(get_width() * (float(bit_depth) / 8.0f)));
+		data = std::shared_ptr<uint8_t>(new uint8_t[get_height() * stride]());
+	}
+
+	size_t Image::get_width() const {
```
anne-gropler: I am not sure if we agreed on under_score for variable names. I thought we use camelCase?

get/set: if it's a getter, just name it width(). At least that's a naming convention that I follow a lot, but it's optional. We might discuss that in our project group, if you want.

source/llassetgen/source/DistanceTransform.cpp
```diff
@@ -1,300 +1,162 @@
...
-        input.reset(new InputType[size]);
-        if (clear) memset(input.get(), 0, size);
-    }
-
-    bool DistanceTransform::inputAt(DimensionType offset) {
-        assert(offset < width * height && input.get());
-        return (input[offset / 8] >> (offset % 8)) & 1;
+    template<typename PixelType, bool flipped, bool invalidBounds>
+    PixelType DistanceTransform::getPixel(PositionType pos) {
+        const Image& image = std::is_same<PixelType, InputType>::value ? input : output;
```
Blaidd-Drwg: Is it idiomatic to return different data depending on the template type (in this case, pixels from different images)? I found it pretty confusing at first.

Lichtso: The alternative is to use a even less obvious types like bool to switch between them.
Thus, I thought reading InputType / OutputType is better than just true / false.

Blaidd-Drwg: It should at least be documented. Also, it isn't needed in `setPixel` since AFAIK we never modify the input, only the output.

Blaidd-Drwg: An alternative would be having two functions `getInputPixel` and `getOutputPixel` which pass the correct image as an argument to `getPixel`, though I don't know if the additional function call would affect performance.

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,367 @@
...
+    template LLASSETGEN_API uint16_t Image::getPixel<uint16_t>(const Vec2<size_t> pos) const;
+    template LLASSETGEN_API uint8_t Image::getPixel<uint8_t>(const Vec2<size_t> pos) const;
+
+    template<typename pixelType>
+    void Image::setPixel(const Vec2<size_t> pos, pixelType in) const {
+        assert(isValid(pos));
+        Vec2<size_t> offset = min + pos;
+        if (bitDepth <= 8) {
+            uint8_t mask = 0xFF;
+            uint8_t in_byte = static_cast<uint8_t>(*reinterpret_cast<uint32_t*>(&in));
```
Blaidd-Drwg: In [DistanceTransform.cpp#L55](https://github.com/hpicgs/openll-asset-generator/blob/bitmap_class/source/llassetgen/source/DistanceTransform.cpp#L55) this function is called either with `in` set to 0 or infinity. In both cases `in_byte` would be 0.

source/llassetgen/source/DistanceTransform.cpp
Lichtso: I combined them to avoid code redundancy

Blaidd-Drwg: You'd be adding a one-liner for `getInputPixel` and a two-liner for `getOutputPixel`. As far as redundancy is concerned, I think it's a negligible trade-off for the clearer interface.

source/llassetgen/source/DistanceTransform.cpp
```diff
@@ -1,300 +1,166 @@
...
-    }
-
-    DistanceTransform::OutputType& DistanceTransform::outputAt(PositionType pos) {
-        assert(pos.x < width && pos.y < height && output.get());
-        return output[pos.y * width + pos.x];
+    template<typename PixelType, bool flipped, bool invalidBounds>
+    PixelType DistanceTransform::getPixel(PositionType pos) {
+        const Image& image = std::is_same<PixelType, InputType>::value ? input : output;
+        if(invalidBounds && !image.isValid(pos))
+            return 0;
```
Blaidd-Drwg: This makes every pixel outside of the bounds black in the eyes of the distance transform. Since we render black characters on white background, distance fields have a white border around them.
testfontatlas_rgb

Blaidd-Drwg: Changing the value to 1 fixes the problem when DeadReckoning is used, though I don't know if this could break `getPixel<OutputType>`. ParabolaEnvelope creates a white border regardless of the value.

Lichtso: The reason for the `TODO` at the end of the line
`(value >= std::numeric_limits<InputType>::max()/2) : value; // TODO`
was that this check should not happen in DistanceTransform but in Image.
The bit depth reduction was not implemented back then, thus this ugly solution.

Lichtso: So we have to decide what black / white interpretation we want and use a consistent one and remove this discretization here.

Blaidd-Drwg: I see. Could you move the check to Image then?

Blaidd-Drwg: Assuming we choose the current b/w interpretation, do you have an idea how to fix ParabolaEnvelope?

Lichtso: Doesn't TTF use 0 / black for empty and 1 / white for set?
Why would you invert the input, it only costs computation power and time...
Please keep in mind that the output can simply be inverted independent of the input by flipping black / white and min / max parameters of Image::exportPng().

source/llassetgen/source/Image.cpp
```diff
@@ -36,8 +36,16 @@ namespace llassetgen {
...
-    Image Image::view(Vec2<size_t> _min, Vec2<size_t> _max) {
-        return Image(_min, _max, stride, bitDepth, data);
+    Image Image::view(Vec2<size_t> outerMin, Vec2<size_t> outerMax, size_t padding) {
+        assert(outerMax.x <= getWidth() && outerMax.y <= getHeight());
+        Vec2<size_t> vec(padding, padding), innerMin = min + outerMin + vec, innerMax = min + outerMax - vec;
+        if(padding > 0) {
+            fillRect(outerMin, Vec2<size_t>(innerMax.x, innerMin.y));
+            fillRect(Vec2<size_t>(innerMax.x, outerMin.y), Vec2<size_t>(outerMax.x, innerMax.y));
+            fillRect(Vec2<size_t>(innerMin.x, innerMax.y), outerMax);
+            fillRect(Vec2<size_t>(outerMin.x, innerMin.y), Vec2<size_t>(innerMin.x, outerMax.y));
```
Blaidd-Drwg: Can be made more concise using initializer lists.

source/llassetgen/source/Image.cpp
```diff
@@ -101,17 +109,43 @@ namespace llassetgen {
...
 
+    template LLASSETGEN_API void Image::fillRect<float>(Vec2<size_t> _min, Vec2<size_t> _max, float in) const;
+    template LLASSETGEN_API void Image::fillRect<uint32_t>(Vec2<size_t> _min, Vec2<size_t> _max, uint32_t in) const;
+    template LLASSETGEN_API void Image::fillRect<uint16_t>(Vec2<size_t> _min, Vec2<size_t> _max, uint16_t in) const;
+    template LLASSETGEN_API void Image::fillRect<uint8_t>(Vec2<size_t> _min, Vec2<size_t> _max, uint8_t in) const;
+    template<typename pixelType>
+    void Image::fillRect(Vec2<size_t> _min, Vec2<size_t> _max, pixelType in) const {
+        for(size_t y = _min.y; y < _max.y; y++) {
+            for(size_t x = _min.x; x < _max.x; x++) {
+                setPixel<uint16_t>({x, y}, in);
```
Blaidd-Drwg: Why `uint16_t`?

source/llassetgen/source/Image.cpp
```diff
@@ -36,8 +36,16 @@ namespace llassetgen {
         :min(Vec2<size_t>(0, 0)), max(Vec2<size_t>(width, height)), stride((width  * _bitDepth + 7) / 8),
         bitDepth(_bitDepth), data(new uint8_t[stride * height]), isOwnerOfData(true) { }
 
-    Image Image::view(Vec2<size_t> _min, Vec2<size_t> _max) {
-        return Image(_min, _max, stride, bitDepth, data);
+    Image Image::view(Vec2<size_t> outerMin, Vec2<size_t> outerMax, size_t padding) {
+        assert(outerMax.x <= getWidth() && outerMax.y <= getHeight());
+        Vec2<size_t> vec(padding, padding), innerMin = min + outerMin + vec, innerMax = min + outerMax - vec;
```
Blaidd-Drwg: Could you put these statements on separate lines and give `vec` a more descriptive name like `paddingVec`?

Blaidd-Drwg: Why are `min` and `max` added to `innerMin` and `innerMax` but not to `outerMin` and `outerMax`?

Lichtso (approved):
The most important features are implemented and tests running. Further optimizations will be postponed to another branch.

source/llassetgen/include/llassetgen/DistanceTransform.h
```diff
@@ -1,73 +1,57 @@
...
+        using InputType = uint8_t;
         using OutputType = float;
         using PositionType = Vec2<DimensionType>;
 
        protected:
-        std::unique_ptr<InputType[]> input;
-        std::unique_ptr<OutputType[]> output;
-        template <typename PixelType>
-        void exportPngInternal(png_struct_def*, OutputType, OutputType);
+        template <typename PixelType, bool flipped = false, bool invalidBounds = false>
```
Felerius, line 16: I think the usual C++ convention is to write type and value template parameters the same way (CamelCase in our case). I would find it useful because it differentiates runtime and compile time values.

source/llassetgen/include/llassetgen/DistanceTransform.h
```diff
@@ -1,73 +1,57 @@
...
-        OutputType outputAtClamped(PositionType pos);
-
-        LLASSETGEN_API void importFreeTypeBitmap(FT_Bitmap_* bitmap, DimensionType padding);
-        LLASSETGEN_API void importPng(std::string path);
-        LLASSETGEN_API void exportPng(std::string path, OutputType blackDistance, OutputType whiteDistance,
-                                      DimensionType bitDepth);
-
-        virtual void transform() = 0;
+        const Image& input;
+        const Image& output;
```
Felerius, line 23: Do these need to be public? If the general usage is "create instance" -> "call transform" -> "do something with output" you have access to the images anyway, because you passed them in when creating the instance.

source/llassetgen/include/llassetgen/Image.h
```diff
@@ -0,0 +1,52 @@
...
+        LLASSETGEN_NO_EXPORT Image(Vec2<size_t> _min, Vec2<size_t> _max, size_t _stride, uint8_t _bitDepth,
+                                   uint8_t* _data);
+        LLASSETGEN_NO_EXPORT static uint32_t reduceBitDepth(uint32_t in, uint8_t in_bitDepth, uint8_t out_bitDepth);
+        LLASSETGEN_NO_EXPORT static size_t getFtBitdepth(const FT_Bitmap_& ft_bitmap);
+        LLASSETGEN_NO_EXPORT static void readData(png_struct_def* png, uint8_t* data, size_t length);
+        LLASSETGEN_NO_EXPORT static void writeData(png_struct_def* png, uint8_t* data, size_t length);
+        LLASSETGEN_NO_EXPORT static void flushData(png_struct_def* png);
+
+       public:
+        ~Image();
```
Felerius, line 30: I don't know the rules for implicitly declared copy constructors/assignment operators by heart, but should they be declared we could easily end up doubly deleting `data` (create copy, delete copy, delete original). I'd think to be sure we should be explicit with regards to copy/move (following the rule of five):
```cpp
Image(const Image&) = delete;
Image(Image&&) = default;
Image& operator=(const Image&) = delete;
Image& operator=(Image&&) = default;
```

source/llassetgen/include/llassetgen/Image.h
```diff
@@ -0,0 +1,52 @@
...
+struct FT_Bitmap_;
+struct png_struct_def;
+
+namespace llassetgen {
+    class LLASSETGEN_API Image {
+        Vec2<size_t> min, max;
+        size_t stride;
+        uint8_t bitDepth;
+        uint8_t* data;
+        bool isOwnerOfData;
```
Felerius, line 19: This is true exactly when the image is not a view, right? If so, I would name this `isView`, and note on the `view` method that views are only valid while their original image exists (which wasn't the case when we were using shared pointers).

source/llassetgen/include/llassetgen/Image.h
```diff
@@ -0,0 +1,52 @@
...
+        LLASSETGEN_NO_EXPORT static void writeData(png_struct_def* png, uint8_t* data, size_t length);
+        LLASSETGEN_NO_EXPORT static void flushData(png_struct_def* png);
+
+       public:
+        ~Image();
+        Image(size_t width, size_t height, size_t _bitDepth);
+        Image view(Vec2<size_t> _min, Vec2<size_t> _max, size_t padding = 0);
+        size_t getWidth() const;
+        size_t getHeight() const;
+        size_t getBitDepth() const;
```
Felerius, line 35: The usual naming convention is avoiding a `get` prefix. In the case of bit depth the variable also needs to be renamed, I used something like `bitDepth_` in the packer classes.

source/llassetgen/include/llassetgen/Image.h
```diff
@@ -0,0 +1,52 @@
...
+        bool isValid(Vec2<size_t> pos) const;
+        template <typename pixelType>
```
Felerius, line 37: For consistency this should be `PixelType`.

source/llassetgen/include/llassetgen/Image.h
```diff
@@ -0,0 +1,52 @@
...
+        pixelType getPixel(Vec2<size_t> pos) const;
+        template <typename pixelType>
+        void setPixel(Vec2<size_t> pos, pixelType data) const;
+        template <typename pixelType = uint8_t>
+        void fillRect(Vec2<size_t> _min, Vec2<size_t> _max, pixelType in = 0) const;
+        void clear() const;
+
+        void load(const FT_Bitmap_& ft_bitmap);
+        Image(const std::string& filepath, uint8_t _bitDepth = 0);
```
Felerius, line 46: Should probably be grouped with the other constructors

source/llassetgen/source/DistanceTransform.cpp
```diff
@@ -1,300 +1,166 @@
...
-        assert(pos.x < width && pos.y < height);
-        return inputAt(pos.y * width + pos.x);
+    template <typename PixelType, bool flipped, bool invalidBounds>
+    PixelType DistanceTransform::getPixel(PositionType pos) {
+        const Image& image = std::is_same<PixelType, InputType>::value ? input : output;
+        if (invalidBounds && !image.isValid(pos))
+            return 0;
+        if (flipped)
+            std::swap(pos.x, pos.y);
+        return image.getPixel<PixelType>(pos);
```
Felerius, line 14: This (and also `setPixel`) is very unclear. What would be the downside to having two `getPixel*` methods, both calling a free `getPixel<PixelType, Flipped, InvalidBounds>(pos, image)` function?

Also, replacing `flipped` and `invalidBounds` with two value enums could maybe make the calling code clearer (something like `getOutputPixel<Flipped, DontCheckBounds>` instead of `getOutputPixel<true, false>`)

source/llassetgen/source/DistanceTransform.cpp
```diff
@@ -1,300 +1,166 @@
...
-                    (center && (inputAtClamped({x - 1, y}) != center || inputAtClamped({x + 1, y}) != center ||
-                                inputAtClamped({x, y - 1}) != center || inputAtClamped({x, y + 1}) != center))
-                        ? 0
-                        : std::numeric_limits<OutputType>::infinity();
+                setPixel<OutputType>(pos,
+                    (center && (getPixel<InputType, false, true>({x - 1, y}) != center || getPixel<InputType, false, true>({x + 1, y}) != center ||
+                                getPixel<InputType, false, true>({x, y - 1}) != center || getPixel<InputType, false, true>({x, y + 1}) != center))
+                    ? 0
+                    : std::numeric_limits<OutputType>::infinity()
+                );
```
Felerius, line 58: Probably easier to read if the pixel values are read into variables first.

source/llassetgen/source/DistanceTransform.cpp
```diff
@@ -1,300 +1,166 @@
...
-                }
+            DimensionType end = (next) ? j : j - 1;
+            if (fill)
+                for (DimensionType i = begin; i < end; ++i)
+                    setPixel<OutputType, flipped>(
+                        {i, offset},
+                        square(begin == 0
+                            ? (end - i)                                            // Falling slope
+                            : ((i < (end + begin) / 2) ? i - begin + 1 : end - i)  // Rising and falling slope
+                        )
```
Felerius, line 103: Mostly screwed up by the formatter, but even with hand formatting this could still be split into multiple lines for readability.

source/llassetgen/include/llassetgen/Image.h
```diff
@@ -0,0 +1,52 @@
...
+        size_t getHeight() const;
+        size_t getBitDepth() const;
+        bool isValid(Vec2<size_t> pos) const;
+        template <typename pixelType>
+        pixelType getPixel(Vec2<size_t> pos) const;
+        template <typename pixelType>
+        void setPixel(Vec2<size_t> pos, pixelType data) const;
+        template <typename pixelType = uint8_t>
+        void fillRect(Vec2<size_t> _min, Vec2<size_t> _max, pixelType in = 0) const;
+        void clear() const;
```
Felerius, line 43: Why are these three methods const? While it is legal C++ because the data is behind a pointer, it is very misleading (`const std::vector` also doesn't allow editing values). Is there any specific reasons that these need to stay const?

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,360 @@
...
+    pixelType Image::getPixel(Vec2<size_t> pos) const {
+        assert(isValid(pos));
+        Vec2<size_t> offset = min + pos;
+        if (bitDepth <= 8) {
+            uint8_t byte = data[offset.y * stride + offset.x * bitDepth / 8];
+            size_t bit_pos = offset.x % (8 / bitDepth);
+            byte <<= bit_pos * bitDepth;
+            byte >>= 8 - bitDepth;
+            uint32_t casted = static_cast<uint32_t>(byte);
+            return reinterpret_cast<pixelType&>(casted);
```
Felerius, line 76: I find it way harder to follow whats going on here compared to the pointer version (especially because I've never seen `reinterpret_cast` used with a reference ever before).

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,360 @@
...
+    template LLASSETGEN_API void Image::setPixel<uint8_t>(Vec2<size_t> pos, uint8_t in) const;
+    template <typename pixelType>
+    void Image::setPixel(Vec2<size_t> pos, pixelType in) const {
+        assert(isValid(pos));
+        Vec2<size_t> offset = min + pos;
+        if (bitDepth <= 8) {
+            uint8_t mask = 0xFF;
+            uint8_t in_byte = static_cast<uint8_t>(reinterpret_cast<uint32_t&>(in));
+            mask <<= 8 - bitDepth;
+            mask >>= 8 - bitDepth;
```
Felerius, line 99: `uint8_t mask = ~(0xFF << bitDepth)` would be shorted and also easier to understand in my opinion.

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,360 @@
...
+            }
+        }
+    }
+
+    void Image::clear() const {
+        if(isOwnerOfData) {
+            memset(data, 0, stride * getHeight());
+        } else {
+            fillRect(min, max, 0);
+        }
```
Felerius, line 133: Another reason to rename `isOwnerOfData` to `isView`, or at least document the relation somewhere.

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    void Image::load(const FT_Bitmap& ft_bitmap) {
+        assert(getWidth() == ft_bitmap.width && getHeight() == ft_bitmap.rows && bitDepth == getFtBitdepth(ft_bitmap));
+        if (min.x == 0 && min.y == 0) {
+            assert(stride == static_cast<size_t>(ft_bitmap.pitch));
+            memcpy(data, ft_bitmap.buffer, ft_bitmap.pitch * ft_bitmap.rows);
+        } else if (min.x % 8 == 0) {
+            assert(bitDepth == 1);
```
Felerius, line 142: Is there a reason that this always has to hold, or is this simply not implemented for bitDepth != 1?

source/llassetgen/include/llassetgen/Image.h
```diff
@@ -0,0 +1,52 @@
...
+        Vec2<size_t> min, max;
+        size_t stride;
+        uint8_t bitDepth;
+        uint8_t* data;
+        bool isOwnerOfData;
+
+        LLASSETGEN_NO_EXPORT Image(Vec2<size_t> _min, Vec2<size_t> _max, size_t _stride, uint8_t _bitDepth,
+                                   uint8_t* _data);
+        LLASSETGEN_NO_EXPORT static uint32_t reduceBitDepth(uint32_t in, uint8_t in_bitDepth, uint8_t out_bitDepth);
+        LLASSETGEN_NO_EXPORT static size_t getFtBitdepth(const FT_Bitmap_& ft_bitmap);
```
Felerius, line 24: Should probably be `getFtBitDepth` for consistency (or just `ftBitDepth`).

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    void Image::writeData(png_structp png, png_bytep data, png_size_t length) {
+        png_voidp a = png_get_io_ptr(png);
+        ((std::ostream*)a)->write((char*)data, length);
+    }
+
+    void Image::flushData(png_structp png) {
+        png_voidp a = png_get_io_ptr(png);
+        ((std::ostream*)a)->flush();
```
Felerius, line 181: Should probably be `reinterpret_cast` for all three.

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,360 @@
...
+
+    Image::Image(const std::string& filepath, uint8_t _bitDepth) {
+        std::ifstream in_file(filepath, std::ifstream::in | std::ifstream::binary);
+
+        png_byte pngsig[8];
+        in_file.read((char*)pngsig, 8);
+        if (!in_file.good()) {
+            std::cerr << "could not read from file";
+            abort();
+        }
```
Felerius, line 198: We don't have a good error handling strategy for the library yet, but writing to stdout and aborting is definitely not it.

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,360 @@
...
+        if (png_sig_cmp(pngsig, 0, 8)) {
+            std::cerr << "no PNG file signature";
+            abort();
+        }
+
+        png_structp png = png_create_read_struct(PNG_LIBPNG_VER_STRING, NULL, NULL, NULL);
```
Felerius, line 204: `NULL` -> `nullptr`

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,360 @@
...
+        if (!png) {
+            std::cerr << "failed to create png read struct";
+            abort();
+        }
+
+        png_infop info = png_create_info_struct(png);
+        if (!info) {
+            png_destroy_read_struct(&png, (png_infopp)0, (png_infopp)0);
```
Felerius, line 212: The latter two arguments could also be `nullptr` I think?

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,360 @@
...
+            } else if (color_type == PNG_COLOR_TYPE_RGB ||
+                color_type == PNG_COLOR_TYPE_RGBA ||
+                color_type == PNG_COLOR_TYPE_RGB_ALPHA) {
+                png_set_rgb_to_gray_fixed(png, 1, -1, -1);
+            }
+            /* TODO: all color types:
+            PNG_COLOR_TYPE_GRAY_ALPHA
+            PNG_COLOR_MASK_PALETTE
+            PNG_COLOR_MASK_COLOR
+            PNG_COLOR_MASK_ALPHA*/
```
Felerius, line 253: Should probably be reported as an error somehow, instead of leaving the image in a uninitialized state.

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,360 @@
...
+        }
+        png_read_update_info(png, info);
+        uint8_t channels = png_get_channels(png, info);
+
+        uint8_t png_bitDepth = png_get_bit_depth(png, info);
+        size_t png_stride = (getWidth() * png_bitDepth + 7) / 8;
+
+        std::unique_ptr<png_bytep> row_ptrs(new png_bytep[getHeight()]);
+        std::unique_ptr<uint8_t> multi_channel_data(new uint8_t[getHeight() * png_stride * channels]);
```
Felerius, line 262: Couldn't this be `std::make_unique<png_bytep[]>(getHeight())` (and respectively for `multi_channel_data`)? Would also make assignment easier (`row_ptrs.get()[i]` -> `row_ptrs[i]`).

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,360 @@
...
+            PNG_COMPRESSION_TYPE_BASE,
+            PNG_FILTER_TYPE_BASE);
+
+        png_write_info(png, info);
+
+        if (bitDepth > 8) {
+            png_set_swap(png);
+        }
+
+        // TODO
```
Felerius, line 335: Whats left to do here (if anything)?

source/llassetgen/source/Image.cpp
```diff
@@ -0,0 +1,360 @@
...
+        } else {
+            for (size_t y = 0; y < getHeight(); y++) {
+                png_write_row(png, reinterpret_cast<png_bytep>(&data[y * stride]));
+            }
+        }
+
+        png_write_end(png, nullptr);
+
+        png_destroy_write_struct(&png, &info);
+        out_file.close();
```
Felerius, line 358: Streams are implicitly closed when they go out of scope (same goes for the importing constructor).

source/tests/llassetgen-tests/Image.cpp
```diff
@@ -0,0 +1,129 @@
...
+#include <gmock/gmock.h>
+#include <llassetgen/llassetgen.h>
+
+#include <fstream>
+
+using namespace llassetgen;
+
+// current working directory: openll-asset-generator/build/source/tests
+std::string test_source_path = "../../../source/tests/llassetgen-tests/testfiles/";
+std::string test_destination_path = "../../";
```
Felerius, line 13: This relies on the location of the cmake build dir compared to the source directory, making it hard to run tests while developing (especially with an IDE). Could we maybe pass the path to the source directory in using CMake (cmake variable -> c++ define -> here)?

In a perfect world we would also use the temp dir as the destination, or just clean up the files after the tests.

source/tests/llassetgen-tests/Image.cpp
```diff
@@ -0,0 +1,129 @@
...
+    char letter = 'J';
+
+    FT_Error face_created = FT_New_Face(freetype, font_file.c_str(), 0, &face);
+    EXPECT_EQ(face_created, 0);
+
+    FT_Error pixel_size_set = FT_Set_Pixel_Sizes(face, 64, 64);
+    EXPECT_EQ(pixel_size_set, 0);
+
+    FT_Error glyph_loaded = FT_Load_Glyph(face, FT_Get_Char_Index(face, letter), FT_LOAD_RENDER | FT_LOAD_MONOCHROME);
+    EXPECT_EQ(glyph_loaded, 0);
```
Felerius, line 29: If the only thing the test is doing should the condition not be met is crashing then these should be `ASSERT_EQ`s

source/tests/llassetgen-tests/Image.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+    size_t padding = 5;
+    Image canvas(face->glyph->bitmap.width + padding * 2, face->glyph->bitmap.rows + padding * 2, 1);
+    auto x = face->glyph->bitmap.width + padding * 2;
+    auto y = face->glyph->bitmap.rows + padding * 2;
+    Image glyph_view = canvas.view({0, 0}, {x, y}, padding);
+    glyph_view.load(face->glyph->bitmap);
+
+    Image invert_view = glyph_view.view(Vec2<size_t>(10, 10), Vec2<size_t>(20, 20));
```
Felerius, line 38: These vectors (and many others in the tests) could be written as `{x, y}`.

source/tests/llassetgen-tests/Image.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+class DistanceTransformTest : public testing::Test {};
+
+TEST_F(DistanceTransformTest, DeadReckoning) {
+    Image input(test_source_path + "Helvetica.png", 1),
+          output(input.getWidth(), input.getHeight(), sizeof(DistanceTransform::OutputType) * 8);
+    std::unique_ptr<DistanceTransform> dt(new DeadReckoning(input, output));
+    dt->transform();
+    output.exportPng<DistanceTransform::OutputType>(test_destination_path + "DeadReckoning.png", -20, 50);
+    EXPECT_EQ(1, 1);
```
Felerius, line 119: Is this required by googletest?

PR #35, opened 2018-01-09
Add simple, fast bin packing method

Felerius:
Adds an implementation of the shelf next-fit bin packing method. This produces inefficient packings, but is very fast (only requires O(n) operations where n is the number of rectangles to pack).

Besides packing into a fixed size atlas it also implements packing with automatically selecting an atlas size. This works by finding the smallest power-of-two atlas size that can hold all rectangles, and iteratively increasing it if packing fails.

The interface for the two packing methods is also designed so that it should work for all bin packing methods (except maybe for optional extra parameters).


source/llassetgen/include/llassetgen/Packing.h
```diff
@@ -0,0 +1,176 @@
...
+    struct Packing {
+        Vec2<PackingSizeType> atlasSize{};
+        std::vector<Rect<PackingSizeType>> rects{};
+
+        Packing() = default;
+
+        explicit Packing(Vec2<PackingSizeType> _atlasSize);
+    };
+
+    namespace impl_packing {
```
anne-gropler: Why underscore? For all variables you used camelCase. Please be consistent. Or is there any style rule that we agreed upon which I forgot? :D Is namespace naming an exception?

anne-gropler:
Can't compile:
in function "expectSuccessfulValidPacking" the "Packing packing{atlasSize}" is not linked, as I understood this error pasted below.

writing LLASSETGEN_API in Packing.h:27 before "explicit Packing(...)" seems to fix that.

Error	LNK2019	unresolved external symbol "public: __cdecl llassetgen::Packing::Packing(struct llassetgen::Vec2<unsigned int>)" (??0Packing@llassetgen@@QEAA@U?$Vec2@I@1@@Z) referenced in function "protected: void __cdecl PackingTest::expectSuccessfulValidPacking(class std::vector<struct llassetgen::Vec2<unsigned int>,class std::allocator<struct llassetgen::Vec2<unsigned int> > > const &,struct llassetgen::Vec2<unsigned int>,bool)" (?expectSuccessfulValidPacking@PackingTest@@IEAAXAEBV?$vector@U?$Vec2@I@llassetgen@@V?$allocator@U?$Vec2@I@llassetgen@@@std@@@std@@U?$Vec2@I@llassetgen@@_N@Z)

source/tests/llassetgen-tests/Packing.cpp
```diff
@@ -0,0 +1,112 @@
...
+    /**
+     * Expect a fixed size packing to succeed and validate the result.
+     */
+    void expectSuccessfulValidPacking(const std::vector<Vec>& rectSizes, Vec atlasSize, bool allowRotations);
+
+    /**
+     * Expects a variable size packing to return a valid result.
+     */
+    void expectValidPacking(const std::vector<Vec>& rectSizes, bool allowRotations) {
+        validatePacking(run(rectSizes, allowRotations), rectSizes, allowRotations);
```
anne-gropler, line 32: Why is this function defined here and not in lines 72ff, as the other functions?

source/tests/llassetgen-tests/Packing.cpp
```diff
@@ -0,0 +1,112 @@
...
+    Packing packing{atlasSize};
+    ASSERT_TRUE(run(rectSizes, allowRotations, packing));
+    EXPECT_EQ(atlasSize, packing.atlasSize);
+    validatePacking(packing, rectSizes, allowRotations);
+}
+
+void PackingTest::validatePacking(Packing packing, const std::vector<Vec>& rectSizes, bool allowRotations) {
+    EXPECT_EQ(rectSizes.size(), packing.rects.size());
+
+    for (size_t i = 0; i < packing.rects.size(); i++) {
```
anne-gropler, line 84: optimization: if packing.rects.size() doesn't change while looping, you can store the size in a new variable (e.g. "auto size = packing.rects.size();"). This way,

(1) the size-function is not called for every iteration.
(2) the reader knows that the size doesn't change while looping.

anne-gropler:
Missing documentation for this PR: How can I try your code? (using cmd line parameters, using a certain function, etc.?)

source/llassetgen/source/Packing.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    namespace impl_packing {
+        bool ShelfNextFitPacker::packNext(Vec2<PackingSizeType> rectSize) {
+            if (currentShelfSize.x + rectSize.x > packing.atlasSize.x) {
+                usedHeight += currentShelfSize.y;
+                if (rectSize.x > packing.atlasSize.x) {
+                    return false;
+                }
+
+                currentShelfSize = {0, 0};
```
Blaidd-Drwg: Could use `openNewShelf` here as well

Felerius:
The symbol visibility is fixed now, I forgot to export it when I changed the way the two packing functions interacted in fda8114.

With regards to documentation: the packing header only contains two public functions (`shelfPackAtlas` and `shelfPackFixedSizeAtlas`), which I think are relatively well documented. To test the code in the context of this PR the easiest way would be to change some values in the test code, or to write a simple command line program using those two functions.

source/tests/llassetgen-tests/Packing.cpp
Felerius: I tend to write very short functions inside the class body just to keep the code shorter, but I don't have really have a strong opinion on what the best style for this is.

source/tests/llassetgen-tests/Packing.cpp
Felerius: I'm pretty sure almost every compiler will optimize that function call away. Even so, this is test code, so I'd prefer it to be concise and readable over perfectly optimized (and with the current tests we really shouldn't have any performance issues).

source/llassetgen/source/Packing.cpp
Felerius: Using it in both methods was the very reason I factored that method out, but I must have forgotten to change it here. Fixed in the newest commit, thanks!

source/llassetgen/include/llassetgen/Packing.h
Felerius: The namespace was originally named `impl`, but I changed it to prevent potential name clashes should we use such a namespace in other headers and didn't really thought much about the name. I now think that that is probably just overcautious, and that we should just stick with `impl` or `detail`. The most important part is that the name makes clear that the contents of the namespace are implementation details and not part of the libraries interface.

Felerius: `impl` and `detail` is also just the naming scheme I know from other C++ libraries, not sure if there is any best/most common choice here.

source/tests/llassetgen-tests/Packing.cpp
anne-gropler: I understand your reasoning.
The best style though is Consistency. If it were just one function inside expectValidPacking (like "run"), then it would be okay for me since it's only "forwarding". This function might be short written, but there is a lot that happens here.
When would it be "long enough" to define that functions at 72ff? There is no rule for that, rather opinions.
I'd really like you to be consistent.
Maybe have another opinion from @Blaidd-Drwg?

source/llassetgen/source/Packing.cpp
```diff
@@ -0,0 +1,88 @@
+#include <llassetgen/Packing.h>
+
+#include <algorithm>
+#include <tuple>
+
+namespace llassetgen {
+    Packing::Packing(Vec2<PackingSizeType> _atlasSize) : atlasSize{_atlasSize} {}
+
+    namespace impl_packing {
```
anne-gropler: Why underscore and not camelCase?
(If we agreed on this style, I just forgot...)

source/llassetgen/include/llassetgen/Geometry.h
```diff
@@ -20,6 +22,14 @@ namespace llassetgen {
         bool operator!=(const Vec2& other) const;
     };
 
+    template <class T>
+    struct Rect {
+        Vec2<T> position{0, 0};
+        Vec2<T> size{0, 0};
+
+        Rect(Vec2<T> _position, Vec2<T> _size);
```
scheibel: Seems like unnecessary copies.

source/llassetgen/include/llassetgen/Packing.h
```diff
@@ -0,0 +1,176 @@
...
+    struct Packing {
+        Vec2<PackingSizeType> atlasSize{};
+        std::vector<Rect<PackingSizeType>> rects{};
+
+        Packing() = default;
+
+        LLASSETGEN_API explicit Packing(Vec2<PackingSizeType> _atlasSize);
+    };
+
+    namespace impl_packing {
```
scheibel: I suggest moving those internals away from the public header.

source/llassetgen/include/llassetgen/Packing.h
```diff
@@ -0,0 +1,176 @@
...
+        constexpr int assertCorrectIteratorType() {
+            using IterTraits = typename std::iterator_traits<Iter>;
+            using IterRefType = typename IterTraits::reference;
+            using IterCategory = typename IterTraits::iterator_category;
+            static_assert(std::is_convertible<IterRefType, const Vec2<PackingSizeType>&>::value,
+                          "Packing input must be convertible to const Vec2<PackingSizeType>&");
+            static_assert(std::is_base_of<std::forward_iterator_tag, IterCategory>::value,
+                          "Iterator must be a ForwardIterator");
+
+            // C++11 doesn't allow void as the return type of constexpr functions
```
scheibel, line 41: Why even wrapping it in a function? `static_assert` is evaluated in class contexts, too.

source/llassetgen/source/Packing.cpp
```diff
@@ -0,0 +1,88 @@
...
+            if (usedHeight + rectSize.y > packing.atlasSize.y) {
+                return false;
+            }
+
+            store(rectSize);
+            return true;
+        }
+
+        bool ShelfNextFitPacker::packNextRotatable(Vec2<PackingSizeType> rectSize) {
+            PackingSizeType minSide, maxSide;
```
scheibel, line 27: We discourage multiple declarations per line.
Further, we discourage declaration without definition / initialization.

source/llassetgen/source/Packing.cpp
```diff
@@ -0,0 +1,88 @@
...
+            currentShelfSize = {0, 0};
+        }
+
+        void ShelfNextFitPacker::store(Vec2<PackingSizeType> rectSize) {
+            packing.rects.push_back({{currentShelfSize.x, usedHeight}, rectSize});
+            currentShelfSize.x += rectSize.x;
+            currentShelfSize.y = std::max(currentShelfSize.y, rectSize.y);
+        }
+
+        unsigned int ceilLog2(uint64_t i) {
```
scheibel: Isn't there an even faster implementation using bitshifts or search for the highest-set bit?
The next power of two of `110110` (6 digits) is `1000000` (7 digits)?

source/llassetgen/source/Packing.cpp
```diff
@@ -0,0 +1,88 @@
...
+            uint64_t powerOfTwo = 1;
+            while (powerOfTwo < i) {
+                powerOfTwo *= 2;
+                exponent++;
+            }
+
+            return exponent;
+        }
+
+        Vec2<PackingSizeType> nextLargerAtlasSize(Vec2<PackingSizeType> previous) {
```
scheibel: Unnecessary copies of input parameter.

source/llassetgen/source/Packing.cpp
Felerius: Copying my comment from your older comment (which Github unfortunately hides because I changed something 2 lines above):

The namespace was originally named `impl`, but I changed it to prevent potential name clashes should we use such a namespace in other headers and didn't really thought much about the name. I now think that that is probably just overcautious, and that we should just stick with `impl` or `detail`. The most important part is that the name makes clear that the contents of the namespace are implementation details and not part of the libraries interface.

`impl` and `detail` is also just the naming scheme I know from other C++ libraries, not sure if there is any best/most common choice here.

anne-gropler: ah, thanks!

source/llassetgen/include/llassetgen/Packing.h
Felerius: I factored this out of the two `shelfPack*` functions to avoid code duplication. Which class context would you move this too? I tried to keep `ShelfNextFitPacker` a non-templated class so that this doesn't become a header-only part of an otherwise not header-only library.

Also, having a function constraints the type aliases to where they are needed.

source/llassetgen/source/Packing.cpp
Felerius: The standard bitshift version doesn't really work here, because it calculates the next power of two and not the logarithm. There is a method which basically does binary search for the highest set bit (something like [this](https://stackoverflow.com/a/15327567)), and there's also builtins for gcc, clang and msvc, which compile down to a single instruction (count leading zeros). Do you think it is worth to do compiler based intrinsic selection, or is the binary search one fast enough?

source/llassetgen/include/llassetgen/Packing.h
Felerius: Good idea, but there is one small problem: Several of the internal functions use `PackingSizeType` in their definitions, which is not an internal type alias and therefore should probably stay in `Packing.h`. Besides placing the alias before the include, my only idea would be to factor it out into a third header. Or just getting rid of the alias all together.

source/llassetgen/source/Packing.cpp
Felerius: Changed to `namespace internal`.

source/llassetgen/source/Packing.cpp
Felerius: Fixed

source/llassetgen/include/llassetgen/Geometry.h
Felerius: Fixed

source/llassetgen/source/Packing.cpp
Felerius: I refactored the code to avoid the `rotated` variable being left uninitialized. I personally prefer `std::tie` over manual tuple unpacking (which would avoid uninitialized variables), at least without C++17/structured bindings.

source/llassetgen/source/Packing.cpp
Felerius: Binary search method implemented in 701414d

Felerius:
@anne-gropler @Blaidd-Drwg Could one of you have a quick look over the two new commits (especially the bit magic `ceilLog2` implementation)? Other than that I think the PR is ready to be merged.

anne-gropler (approved):
I've never been good in understanding bit magic, but the way you commented it, it seems understandable. If your test works, then fine! :)

Issue #36, opened 2018-02-21
Interactive Rendering


anne-gropler:
Develope an application which uses llassetgen and displays GUI for interacting with the rendered image.

- interactive colouring:
    - [x] font color
    - [x] background color 

- 3D:
    - [x] pan
    - [x] zoom
    - [x] rotate

- other:
    - [x] button/dialog to load font
    - [x] button/dialog to export fontatlas and correspondent data
    - [x] dropdown list to choose Distance Transform Algorithm
    - [x] dropdown list to choose Anti-Aliasing / Supersampling
    - [x] toggle between Distance Field and Rendering Glyphs
    - [ ] ~~visual comparison for different choices mentioned above~~

anne-gropler:
interactive colouring done in PR #37

anne-gropler:
panning, zooming and rotating done in PR #44

anne-gropler:
super sampling done in PR #48.

anne-gropler:
toggle between Distance Field and Rendering Glyphs done in PR #49

anne-gropler:
Following GUI Elements are part of PR #54 :

- button/dialog to load font
- button/dialog to export fontatlas and correspondent data
- dropdown list to choose Distance Transform Algorithm

At this point, there won't be any parallel visual comparisons between different parameter constellations.
"visual comparison for different choices mentioned above" is obsolet.

anne-gropler:
There is no "button/dialog to load font", because the font is loaded automatically using the font name.

PR #37, opened 2018-02-21
First rendering

anne-gropler:
This pull request implements some features from Issue #36, namely the interactive colouring (font color, background color).
Starting the application llassetgen-rendering, you shoud see something like this (different colours).
pr

Why did I separately write the classes `Window `and `WindowQt`, instead of merging them into one?
That's because glbinding is not compatible with gl.h, which might get included by some Qt-classes. Merging `Window `and `WindowQt` throws the error "glbinding is not compatible with gl.h" located in glbinding/nogl.h. To fix this, the GL-rendering we separate the gl from the Qt using those 2 classes, which is encouraged by glbinding authors.

HELP WANTED: Please tell me if the shaders and the fontatlas are loaded from the correct path on other OS (file structure might be different).

WIP: fix CI dependencies.


anne-gropler:
@scheibel 
Travis can't build on Mac OS because homebrew installs globjects v1 instead of v2.
apt get on linux installs v2.
Appveyour can't build on windows because there is no NuGet nor Chocolatey for glbinding and globjects.

Independently from that, I'd appreciate reviews for this PR.

.travis.yml
```diff
@@ -14,6 +14,10 @@ compiler:
 env:
   global:
     - CMAKE_OPTIONS="-DOPTION_BUILD_EXAMPLES=On"
+    - PATH="/usr/local/opt/qt/bin:$PATH"
+    - PATH="/usr/local/opt/glbinding:$PATH"
+    - PATH="/usr/local/opt/globjects:$PATH"
```
Felerius: I would put these all into one line, otherwise the list format suggests that three separate variables are being set

.travis.yml
```diff
@@ -24,9 +28,13 @@ matrix:
...
 before_install:
-  - if [ $TRAVIS_OS_NAME == linux ]; then sudo apt-get install -qq cmake3 cppcheck clang-tidy-3.8 libfreetype6-dev libpng-dev; fi
+  - if [ $TRAVIS_OS_NAME == linux ]; then sudo apt-add-repository ppa:cginternals/ppa -y; fi
+  - if [ $TRAVIS_OS_NAME == linux ]; then sudo apt-add-repository ppa:cginternals/backports-ppa -y; fi
+  - if [ $TRAVIS_OS_NAME == linux ]; then sudo apt-get update -qq; fi
+  - if [ $TRAVIS_OS_NAME == linux ]; then sudo apt-get install -qq cmake3 cppcheck clang-tidy-3.8 libfreetype6-dev libpng-dev qt5-default libglm-dev libglbinding-dev libglobjects-dev; fi
   # libpng is already installed by default
-  - if [ $TRAVIS_OS_NAME == osx ]; then brew update && brew install cppcheck freetype; fi
+  - if [ $TRAVIS_OS_NAME == osx ]; then brew tap cginternals/homebrew-cginternals; fi
+  - if [ $TRAVIS_OS_NAME == osx ]; then brew update && brew install cppcheck freetype qt5 glm glbinding globjects; export CMAKE_PREFIX_PATH="$CMAKE_PREFIX_PATH;$(brew --prefix qt);$(brew --prefix glbinding);$(brew --prefix globjects)"; fi
```
Felerius: I think this could be turned into a multi line string like so (to avoid repeating the if statement making the whole thing more readable):
```yaml
- |-
  if [ ... ]; then
    ...;
    ...;
  fi
```

The `|` starts a multi line string, and the additional `-` cuts out trailing newlines from the string (not necessarily needed).

.travis.yml
```diff
@@ -24,9 +28,13 @@ matrix:
     compiler: gcc
 
 before_install:
-  - if [ $TRAVIS_OS_NAME == linux ]; then sudo apt-get install -qq cmake3 cppcheck clang-tidy-3.8 libfreetype6-dev libpng-dev; fi
+  - if [ $TRAVIS_OS_NAME == linux ]; then sudo apt-add-repository ppa:cginternals/ppa -y; fi
+  - if [ $TRAVIS_OS_NAME == linux ]; then sudo apt-add-repository ppa:cginternals/backports-ppa -y; fi
+  - if [ $TRAVIS_OS_NAME == linux ]; then sudo apt-get update -qq; fi
+  - if [ $TRAVIS_OS_NAME == linux ]; then sudo apt-get install -qq cmake3 cppcheck clang-tidy-3.8 libfreetype6-dev libpng-dev qt5-default libglm-dev libglbinding-dev libglobjects-dev; fi
```
Felerius: As with below, the repeated if statements make this kind of hard to read, but we can update this to use the alternative way of specifying apt dependencies once #39 with 7d13192 gets merged.

appveyor.yml
```diff
@@ -45,13 +45,23 @@ install:
...
 
+- if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2015" ( set Qt5Core_DIR=C:\Qt\5.9.2\msvc2015_64\lib\cmake\Qt5Core)
+- if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2015" ( set Qt5Gui_DIR=C:\Qt\5.9.2\msvc2015_64\lib\cmake\Qt5Gui)
+- if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2015" ( set Qt5OpenGL_DIR=C:\Qt\5.9.2\msvc2015_64\lib\cmake\Qt5OpenGL)
+- if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2015" ( set Qt5Widgets_DIR=C:\Qt\5.9.2\msvc2015_64\lib\cmake\Qt5Widgets)
+
+- if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2017" ( set Qt5Core_DIR=C:\Qt\5.9.2\msvc2017_64\lib\cmake\Qt5Core)
+- if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2017" ( set Qt5Gui_DIR=C:\Qt\5.9.2\msvc2017_64\lib\cmake\Qt5Gui)
+- if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2017" ( set Qt5OpenGL_DIR=C:\Qt\5.9.2\msvc2017_64\lib\cmake\Qt5OpenGL)
+- if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2017" ( set Qt5Widgets_DIR=C:\Qt\5.9.2\msvc2017_64\lib\cmake\Qt5Widgets)
```
Felerius: I don't know the appveyor yaml syntax very well, but if there is a way do to multiline if statements that would make this way more readable.

source/llassetgen-rendering/CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+#
+# Sources
+#
+
+set(CMAKE_AUTOMOC TRUE)
+
+set(sources
+    main.cpp
+	
+	WindowQt.cpp
```
Felerius: There is a lot of mixing of tabs and spaces in this file. I think we use spaces in every other `CMakeLists.txt` and should probably spay consistent.

source/llassetgen-rendering/main.cpp
```diff
@@ -0,0 +1,383 @@
...
+#include <globjects/Shader.h>
+#include <globjects/Texture.h>
+#include <globjects/VertexArray.h>
+#include <globjects/VertexAttributeBinding.h>
+
+#include <llassetgen/llassetgen.h>
+
+#include "WindowQt.h"
+
+using namespace gl;
```
Felerius, line 48: I don't know how many types in this file are from the `gl` namespace, but if its not too many a few `using gl::typename` lines would do the same and not pull in all members of the namespace.

source/llassetgen-rendering/main.cpp
```diff
@@ -0,0 +1,383 @@
...
+ * It handles the GL context and loads the shader from file.
+ * Some functions are connected to the GUI using Qt signals and slots.
+ */
+class Window : public WindowQt {
+   public:
+    explicit Window(const QSurfaceFormat &format) : WindowQt(format) {
+        m_samplerIndex = 0;
+        m_texture = nullptr;
+        m_backgroundColor = glm::vec4(1.f, 1.f, 1.f, 1.f);
+        m_fontColor = glm::vec4(0.f, 0.f, 0.f, 1.f);
```
Felerius: As far as i know it is preferred to initialize these directly where the member variable is declared (i.e. `glm::vec4 m_fontColor {0.f, 0.f, 0.f, 1.f}` at the end of the class)

source/llassetgen-rendering/main.cpp
```diff
@@ -0,0 +1,383 @@
...
+    std::unique_ptr<globjects::File> m_fragmentShaderSource;
+    std::unique_ptr<globjects::AbstractStringSource> m_fragmentShaderTemplate;
+    std::unique_ptr<globjects::Shader> m_fragmentShader;
+    std::unique_ptr<globjects::VertexArray> m_vao;
+
+    std::unique_ptr<globjects::Texture> m_texture;
+
+    glm::vec4 m_backgroundColor;
+    glm::vec4 m_fontColor;
+    int m_samplerIndex;
```
Felerius: Is the `m_` prefix there to stay consistent with Qt or any other library? Otherwise I would use the same style we use everywhere else.

source/llassetgen-rendering/main.cpp
```diff
@@ -0,0 +1,383 @@
...
+
+        globjects::debug() << "Using global OS X shader replacement '#version 140' -> '#version 150'" << std::endl;
+#endif
+
+        // get glyph atlas
+
+        // TODO load using own png loader instead of Qt (blocked: wait for this feature to be merged into master, then
+        // pull). TODO: Make sure this file structure is the same on other OS
+        auto path = QApplication::applicationDirPath();
+        auto *image = new QImage(path + "/../../data/llassetgen-rendering/testfontatlas_rgb.png");
```
Felerius: If we want to include a default font atlas long-term, we should probably include it in the executable or figure out how to install into and reference the standard directories (stuff like this would probably go into `/usr/[local/]share` on Linux and macOS).

source/llassetgen-rendering/main.cpp
```diff
@@ -0,0 +1,383 @@
...
+
+        m_cornerBuffer.reset(nullptr);
+        m_program.reset(nullptr);
+        m_vertexShaderSource.reset(nullptr);
+        m_vertexShaderTemplate.reset(nullptr);
+        m_vertexShader.reset(nullptr);
+        m_fragmentShaderSource.reset(nullptr);
+        m_fragmentShaderTemplate.reset(nullptr);
+        m_fragmentShader.reset(nullptr);
+        m_vao.reset(nullptr);
```
Felerius: `variable = nullptr` does the same thing and is a bit more clear in my opinion.

source/llassetgen-rendering/main.cpp
```diff
@@ -0,0 +1,383 @@
...
+        int green = value.toInt();
+        m_fontColor.g = green / 255.f;
+        paint();
+    }
+
+    virtual void fontColorBChanged(QString value) override {
+        int blue = value.toInt();
+        m_fontColor.b = blue / 255.f;
+        paint();
+    }
```
Felerius: Is there a way to deduplicate these methods? As it stands, it is basically the same code copy pasted 6 times. Maybe something like:
```cpp
void applyColorChange(float& color, QString value) {
    color = value.toInt() / 255.f;
    paint();
}

virtual void fontColorBChanged(QString value) override {
    applyColorChange(m_fontColor.b, value);
}
```

source/llassetgen-rendering/main.cpp
```diff
@@ -0,0 +1,383 @@
...
+
+    backgroundColorLayout->addRow(labelG, backgroundG);
+
+    // Background Color BLUE
+    auto *backgroundB = new QLineEdit();
+    auto *labelB = new QLabel("B");
+    labelB->setMaximumWidth(90);
+    backgroundB->setValidator(colorValidator);
+    backgroundB->setPlaceholderText("255");
+    backgroundB->setMaximumWidth(45);
```
Felerius: Could probably made shorter by making a factory methods for color fields

Felerius (approved):
I left a few comments (mostly nitpicky code style related stuff, feel free to ignore some things if you think they are unimportant), but the code generally looks good. I didn't get to test the code because I don't have time to build globjects myself right now (I'm using macOS)

.travis.yml
anne-gropler: I'd do that, too.. actually I was confused about the ":", I expected a ";", so I just copied the line to make sure I won't make syntax mistakes :D I'm not used to linux nor macOS

.travis.yml
anne-gropler: Cool, thanks!

source/llassetgen-rendering/CMakeLists.txt
anne-gropler: Thanks. I didn't see that my Editor used tabs.

source/llassetgen-rendering/main.cpp
anne-gropler: There are currently 20 identifiers that need that namespace, and they all start with gl. So instead of `GL_TEXTURE_2D` I'd need to write `gl::GL_TEXTURE_2D`; or `glClear becomes `gl::glClear`.
So I'd rather use this namespace as it is. 

Well, I am not sure... are 20 identifiers few or many?

source/llassetgen-rendering/main.cpp
anne-gropler: Ok :)

source/llassetgen-rendering/main.cpp
anne-gropler: m_ stands for member variable, it's a coding style we use for years at the CGS, that's why I automatically used it, sorry :D
So, as far as I see, the style in this project is to simply remove the m_?

source/llassetgen-rendering/main.cpp
anne-gropler: This path finds the executable path and works relative from there.
It's not about including a default atlas long-term, but asking the user to specify a path where to load from. Until then, we just load a png as a mock. That's the plan :D

source/llassetgen-rendering/main.cpp
anne-gropler: Since those variables are unique pointers, I'd stick to reset ;)

source/llassetgen-rendering/main.cpp
anne-gropler: yep, good idea, thanks!

source/llassetgen-rendering/main.cpp
anne-gropler: A factory (as in GoF) needs subclasses, I think that's too much here. But I created a method that replaces 3 lines and makes maintenance easier.

source/llassetgen-rendering/main.cpp
Felerius: Oh, ok. If that namespace just contains the usual opengl symbols (which are prefixed with `GL_`, `gl`, etc. anyway), then the `using namespace` is fine.

appveyor.yml
anne-gropler: I tried various multi-lines and none of them work. Sorry :(

anne-gropler:
@scheibel @cgcostume You can start reviewing. The WIP only applies for the Appveyor dependencies.
Observe https://help.appveyor.com/discussions/problems/12635-using-vcpkg-for-glm-etc-cmake-doesnt-find-the-preinstalled-qt-anymore

Blaidd-Drwg:
Re: help wanted: I figure that also depends on where your IDE places the executable. Clion keeps it in `openll-asset-generator/cmake-build-debug`, so I have to delete one `../` from every path to make it work.

anne-gropler:
Thanks, that's good to know! I'll try to find a better way for loading files, but that wil be another PR. Hopefully, I won't need it, since loading from files should be done by our image class.

anne-gropler:
Changed near plane from 0.1 to 0.0001 as @Blaidd-Drwg requested in an extern conversation.

source/llassetgen-rendering/main.cpp
scheibel: Yes, we designed the `gl*` namespaces to include only the official symbols from the OpenGL API of their respective version (namespace `gl` includes all).

PR #38, opened 2018-02-27
Set default symbol visibility to hidden

Felerius:
The related line from cmake-init was commented for some reason. Enabling this should make it easier to detect linker errors on macOS and linux, and results in less bloated interfaces/dynamic libraries.

This also caused some new linker errors in the distance transform code. The solution to that was to export the relevant classes (which MSVC seemingly doesn't require, otherwise we would have gotten errors earlier). This by default exports all class methods, so I excluded the methods that we don't use. On that note, why are all the non exported methods in `DistanceTransform` public? Should they be exported as well? @Lichtso


Lichtso:
Well the `DistanceTransform` interfaces on the master branch are old, the newest are in #32 .
So applying these changes now would only result in a unnecessary merge conflict later.
But I can transfer the changes to the other PR.

PR #39, opened 2018-02-27
Update cmake-init

Felerius:
Applies all but one commits that have been made to cmake-init since we included it. I excluded cginternals/cmake-init@f0140a2 because it changed `.travis.yml` incompatibly and instead picked out the relevant changes into another commit (mostly using another way to install apt dependencies, allowing us to ditch `sudo: required` which hopefully means faster builds).

I also updated `META_CMAKE_INIT_SHA` and marked the relevant `ADAPT` task as done.

~~Based on #38 because it made applying the patch easier.~~
Rebased to not rely on #38 which was closed in favor of having the same changes be part of #32.


.travis.yml
```diff
@@ -23,15 +21,23 @@ matrix:
...
+      - cmake3
+      - cppcheck
+      - clang-tidy-3.8
+      - libfreetype6-dev
+      - libpng-dev
+
 before_install:
-  - if [ $TRAVIS_OS_NAME == linux ]; then sudo apt-get install -qq cmake3 cppcheck clang-tidy-3.8 libfreetype6-dev libpng-dev; fi
   # libpng is already installed by default
   - if [ $TRAVIS_OS_NAME == osx ]; then brew update && brew install cppcheck freetype; fi
```
anne-gropler: Is there a way to make syntax for brew the same as for the apt packages in lines 24-31?

.travis.yml
```diff
@@ -23,15 +21,23 @@ matrix:
   - os: osx
     compiler: gcc
 
+addons:
```
anne-gropler, line 24: Is there a way to use a similar syntax for brew?

anne-gropler: And how do we do this when we need extra ppa?

Felerius: According to travis-ci/travis-ci#5377 there is nothing like this for homebrew yet, unfortunately. For apt, you can ppas to a `sources` list (copied from travises documentation):
```yaml
addons:
  apt:
    sources:
    - ubuntu-toolchain-r-test
    packages:
    - gcc-4.8
    - g++-4.8
```

anne-gropler: I've had a look into it. We can only use white-listed sources.
In my sub-branch I'll need globjects and glbinding, which are not whitelisted. That means, that we need `sudo required` either way.

anne-gropler: > At this moment, we are unable to continue processing the APT sources requests. If you require a new source, please use `sudo: required` in your `.travis.yml`  file.


See https://github.com/travis-ci/apt-source-whitelist

Felerius: I thought glbinding and globjects had official ubuntu/debian packages, or are they too outdated? Anyway, that's unfortunate, but the linux builds will probably still be faster than the macOS builds.

I guess we just revert to manual apt installation when merging your PR? You're changing the `.travis.yml` substantially anyway.

anne-gropler: Ok, fine by me :)

anne-gropler (approved):

PR #41, opened 2018-03-06
Optimize fast packing method

Felerius:
Avoids throwing away partial results when the guessed atlas size is too small.

Downside: Texture atlases might be less square (only height gets increased after initial guess). Should only matter if initial guess for the size is more than 2x off, i.e. the resulting packing has less than 50% space efficiency.

### DIFF
--- a/source/llassetgen/source/Packing.cpp
+++ b/source/llassetgen/source/Packing.cpp
@@ -3,11 +3,21 @@
 #include <algorithm>
 #include <tuple>
 
+using llassetgen::PackingSizeType;
+
+PackingSizeType ceilDiv(PackingSizeType divident, PackingSizeType divisor) {
+    return (divident + divisor - 1) / divisor;
+}
+
 namespace llassetgen {
     Packing::Packing(Vec2<PackingSizeType> _atlasSize) : atlasSize{_atlasSize} {}
 
     namespace internal {
         bool ShelfNextFitPacker::packNext(Vec2<PackingSizeType> rectSize) {
+            return allowRotations ? packNextWithRotations(rectSize) : packNextNoRotations(rectSize);
+        }
+
+        bool ShelfNextFitPacker::packNextNoRotations(Vec2<PackingSizeType> rectSize) {
             if (currentShelfSize.x + rectSize.x > packing.atlasSize.x) {
                 openNewShelf();
                 if (rectSize.x > packing.atlasSize.x) {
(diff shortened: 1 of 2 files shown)


Blaidd-Drwg:
How likely is it that the initial guess is 2x off? Could this realistically happen with real-world data?

Felerius:
Not very likely with real-world data, but it could happen in some edge cases (if some really high rectangles get put into rows with otherwise only very low rectangles, leading to wasted space above the latter). This is because the initial guess is only based on the area of the packed rects.

Something that would definitely make this less likely is sorting the input before processing (for example by width/longer side). This in general improves space efficiency by a lot (according to the "A Thousand Ways to Pack the Bin" paper) but makes packing somewhat slower.

source/llassetgen/source/Packing.cpp
```diff
@@ -16,43 +26,33 @@ namespace llassetgen {
             }
 
             if (usedHeight + rectSize.y > packing.atlasSize.y) {
-                return false;
+                if (allowGrowth) {
+                    PackingSizeType finalHeight = usedHeight + rectSize.y;
+                    auto numDoublings = ceilLog2(ceilDiv(finalHeight, packing.atlasSize.y));
```
Blaidd-Drwg: Given `predictAtlasSize`'s behavior, this should always be 1, right?

source/llassetgen/source/Packing.cpp
```diff
@@ -3,11 +3,21 @@
 #include <algorithm>
 #include <tuple>
 
+using llassetgen::PackingSizeType;
+
+PackingSizeType ceilDiv(PackingSizeType divident, PackingSizeType divisor) {
+    return (divident + divisor - 1) / divisor;
```
Blaidd-Drwg: dividend

source/llassetgen/source/Packing.cpp
```diff
@@ -16,43 +26,33 @@ namespace llassetgen {
...
-            if (upright) {
+            if (currentShelfSize.y >= maxSide && remainingWidth >= minSide) {
                 store({minSide, maxSide});
-            } else {
+            } else if (remainingWidth >= maxSide && remainingHeight >= minSide) {
                 store({maxSide, minSide});
+            } else {
+                openNewShelf();
+                return maxSide > packing.atlasSize.x ? packNextNoRotations({minSide, maxSide})
+                                                     : packNextNoRotations({maxSide, minSide});
```
Blaidd-Drwg: This is in order to let `packNextNoRotations` do the growing if needed, right? Seems a bit convoluted, maybe you could extract the growing into a separate function?

Blaidd-Drwg:
I see. Well I guess it's ok to return sub-par results with the fastest algorithm if the next-fastest option (sorting the input) is such a big improvement with such little cost that it is preferable in nearly every use case anyway.
We'll just have to make sure we discourage using the fastest option in the documentation.

Felerius:
The reason I implemented this packing method was to:
 - have a simple packing method that's quick to implement to establish the interface, and
 - have a packing method that is definitely fast enough for interactive use (previewing in a GUI app).

For both use cases the packing efficiency not very important. My plan is to first implement a more sophisticated "slow" packing method and then see if there is actually a need for a "fast" packing method for normal use, or if the "slow" one is already fast enough.

source/llassetgen/source/Packing.cpp
Felerius: Without rotations yes, with rotations I'm not entirely sure (`predictAtlasSize` does not take rotations into account). I'd rather be defensive here and leave it as is (also allows changes to how we calculate the atlas size without changes being needed here).

source/llassetgen/source/Packing.cpp
Felerius: Thanks, fixed

source/llassetgen/source/Packing.cpp
Felerius: Yeah, in hindsight that was not very clear. I did what you suggested and factored the growth logic out into another method.

Blaidd-Drwg (approved):

PR #43, opened 2018-03-07
Disable adaptive pointer formatting

Felerius:
clang-format derives pointer formatting (`type* name` vs. `type *name`) from the unformatted source file by default. This turns this off in favor of `type* name`.


Blaidd-Drwg (approved):

anne-gropler:
Thanks for this quick fix!

PR #44, opened 2018-03-08
Rendering transforms

anne-gropler:
This pull request implements some features from Issue #36, namely the 3D Transforms (panning, zooming, rotating).
These interactions are important, because they show the glyphs in positions unusual for 2D, but important for readability and rendering results in 3D.

Before merging this PR, PR #37 should be merged (until then, all commits for PR #37 will appear here).
Known Issue: Appveyor and the globjects dependencies.


anne-gropler:
Changed near plane from 0.1 to 0.0001 as @Blaidd-Drwg requested in an extern conversation.

source/llassetgen-rendering/main.cpp
```diff
@@ -77,7 +80,7 @@ class Window : public WindowQt {
         // TODO load using own png loader instead of Qt (blocked: wait for this feature to be merged into master, then
         // pull). TODO: Make sure this file structure is the same on other OS
         auto path = QApplication::applicationDirPath();
-        auto *image = new QImage(path + "/../../data/llassetgen-rendering/testfontatlas_rgb.png");
+        auto* image = new QImage(path + "/../../data/llassetgen-rendering/testfontatlas_DT.png");
```
Felerius, line 83: I probably should have noticed this on the last review already, but how is the allocated `QImage` cleaned up? As far as I can tell it is just used in this method, and never deleted. Wouldn't a stack variable suffice and avoid leaking memory? (Note: This might just be me misunderstanding how Qt handles memory allocation).

source/llassetgen-rendering/main.cpp
```diff
@@ -107,15 +113,29 @@ class Window : public WindowQt {
...
         program->attach(vertexShader.get(), fragmentShader.get());
 
-        cornerBuffer->setData(
-            std::array<glm::vec2, 4>{{glm::vec2(0, 0), glm::vec2(1, 0), glm::vec2(0, 1), glm::vec2(1, 1)}},
-            GL_STATIC_DRAW);
+        float quadW = 1.f;
+        float quadH = quadW * imageH / imageW;
+
+        cornerBuffer->setData(std::array<glm::vec2, 4>{{glm::vec2(0, 0), glm::vec2(quadW, 0), glm::vec2(0, quadH),
+                                                        glm::vec2(quadW, quadH)}},
```
Felerius: I think `glm::vec2(x, y)` could be written as `{x, y}` here (type is already specified in the `std::array` template parameter (unless you find it less readable it less readable of course). Same comment applies for `textureBuffer->setData` a few lines below.

source/llassetgen-rendering/main.cpp
```diff
@@ -179,7 +205,73 @@ class Window : public WindowQt {
...
+        lastMousePos.x = event->x();
+        lastMousePos.y = event->y();
+
+        if (event->button() == Qt::LeftButton) {
+            isPanning = true;
+        } else if (event->button() == Qt::RightButton) {
+            isRotating = true;
+        }
+
+        doneCurrent();
```
Felerius: I don't fully understand what `makeCurrent`/`doneCurrent` is used for (except that it is some OpenGL synchronization stuff), but is it really necessary here if it isn't in `mouseReleaseEvent`?

data/llassetgen-rendering/shader.vert
```diff
@@ -2,13 +2,17 @@
...
 out vec4 color;
 out vec2 v_uv;
 
 void main()
 {
-    gl_Position = vec4(corner * 2.0 - 1.0, 0.0, 1.0);
+    gl_Position = projection * modelView * vec4(corner * 2.0 - 1.0, -2.0, 1.0);
     color = vec4(corner, 0.0, 1.0);
-	v_uv = corner;
+	v_uv = texUV;
```
Felerius: Some tabs/spaces mixing here and in `shader.frag`

source/llassetgen-rendering/WindowQt.cpp
```diff
@@ -127,6 +127,14 @@ bool WindowQt::event(QEvent* event) {
 
 void WindowQt::paintGL() { updateGL(); }
 
+void WindowQt::mousePressEvent(QMouseEvent* /*event*/) {}
+
+void WindowQt::mouseMoveEvent(QMouseEvent* /*event*/) {}
+
+void WindowQt::mouseReleaseEvent(QMouseEvent* /*event*/) {}
+
+void WindowQt::wheelEvent(QWheelEvent* /*event*/) {}
```
Felerius, line 136: I think these are all overrides and could be marked as such.

Felerius (approved):
Left some comments, but looks good to me overall.

Small note (because this is only temporary anyway): Is it intended that the input file `testfontatlas.png` is not part of the repo, while two other PNGs which I don't think are used are?

source/llassetgen-rendering/main.cpp
anne-gropler: QImage is destroyed when going out of scope. :) imageData stays until the texture is "resetted" in deinitializeGL().

source/llassetgen-rendering/main.cpp
anne-gropler: Thanks! This makes the lines even more readable.

source/llassetgen-rendering/main.cpp
anne-gropler: We don't need it in `mouseReleaseEvent` because there we do nothing with OpenGL.
I just had a further look:
(1) I remember that I wrote those calls in mousePressEvent because I had OpenGL-functions there. Since they are gone now, I can delete those calls.
(2) We actually don't need it here in `mouseMoveEvent` because in `paint()` we also have those calls. 

I'll fix this.
Thanks!

data/llassetgen-rendering/shader.vert
anne-gropler: This is already fixed in next PR, but I'll fix it here, too.

source/llassetgen-rendering/WindowQt.cpp
anne-gropler: Yes. They are now marked as such in the header file.

anne-gropler:
@Felerius 

> Small note (because this is only temporary anyway): Is it intended that the input file testfontatlas.png is not part of the repo, while two other PNGs which I don't think are used are?

Yes, you are right. I lost track. I will remove them.

PR #45, opened 2018-03-12
Add MaxRects packer

Felerius:
Adds an implementation of a MaxRects packer (best performing algorithm in the "A Thousand Ways to Pack the Bin" paper, except that we don't use multiple bins aka textures). Also includes a lot of refactoring to separate internal and public header code and avoid code duplication between the two packing algorithms.


source/llassetgen/include/llassetgen/Geometry.h
```diff
@@ -72,9 +79,39 @@ namespace llassetgen {
...
+    bool Rect<T>::operator!=(const Rect& other) const {
+        return !(*this == other);
+    }
+
+    template <class T>
+    bool Rect<T>::contains(const Rect<T>& other) const {
+        auto thisMaxPoint = position + size;
+        auto otherMaxPoint = other.position + other.size;
+        return position.x <= other.position.x && position.y <= other.position.y && thisMaxPoint.x >= otherMaxPoint.x &&
+               thisMaxPoint.y >= otherMaxPoint.y;
```
anne-gropler, line 97: Indentation seems off -- is this just github?

source/llassetgen/include/llassetgen/Packing.h
```diff
@@ -1,173 +1,10 @@
...
-    bool shelfPackFixedSizeAtlas(ForwardIter sizesBegin, ForwardIter sizesEnd, bool allowRotations, Packing& packing) {
-        internal::assertCorrectIteratorType<ForwardIter>();
-
-        internal::ShelfNextFitPacker packer{packing, allowRotations, false};
-        return std::all_of(sizesBegin, sizesEnd,
-                           [&](Vec2<PackingSizeType> rectSize) { return packer.packNext(rectSize); });
-    }
-}
+/** @file
+ * Used to include all packing headers at once.
```
anne-gropler, line 4: So, why is it needed now? To hide that Packing always needs Algorithms and Types? User convenience? (just asking)

Felerius: Just for convenience (only one include needed). The only reason that the public headers needed to be split in two was to break circular references between the public and internal headers, and I wanted to preserve the original header as an "include everything" option.

anne-gropler: That's seems a pretty good idea to me.

source/llassetgen/include/llassetgen/Geometry.h
Felerius: It's aligned with the beginning of the statement on the line above, same as in the `overlaps` method below. It looks a bit weird here because the first line is way longer than the second.

anne-gropler: I mean, it seems that `format` did its job, so it's fine by me, just wanted to know.

anne-gropler: Yes, it was the longer line that was cutted in one of those github views, that's why I asked. When I open that file in my IDE everthing's fine.

source/llassetgen/include/llassetgen/packing/internal/Common.h
```diff
@@ -0,0 +1,158 @@
...
+            });
+            return std::all_of(std::begin(indices), std::end(indices),
+                               [&](size_t i) { return packer.pack(packing.rects[i]); });
+        }
+
+        /**
+         * Create a flexible size packing from given rectangle sizes.
+         *
+         * @tparam Packer
+         *   Packer class. Must provide the following methods:
```
Blaidd-Drwg, line 96: Why not add the required methods to `BasePacker` and assert that `Packer` is a subclass of `BasePacker`?

scheibel: For either run-time member and static functions as well as compile-time functions and variables there are different idioms to ensure subclass interface availability.
* run-time member function: abstract/virtual member functions (subclassing required)
* run-time static functions: immediate/deferred nested member lookup / concepts (subclassing suitable but not required)
* compile-time functions: nested member lookup / concepts (subclassing suitable but not required)

Felerius: I would like to avoid virtual methods, especially because I would have to make one static method non-static. I could move the documentation to `BasePacker`, but that class basically just the common members/methods from `MaxRectsPacker` and `ShelfPacker` factored out into a class, I don't know if that is a better place.

In general though, this packing class interface is completely internal and only needed when defining a new packing algorithm. Also the compiler errors are relatively clear when a method is missing, so I don't think we really need to check at compile time that the methods exist, just describe somewhere what they should do.

Compiler error example:
```
/Users/david/Documents/Uni/adcpp/source/llassetgen/include/llassetgen/packing/internal/Common.h:86:32: error: no member named 'inputSortingComparator' in 'llassetgen::internal::MaxRectsPacker'
                return Packer::inputSortingComparator(packing.rects[i1], packing.rects[i2]);
```

Blaidd-Drwg: Fair enough. Why do you want to avoid virtual methods?

anne-gropler (approved):

Blaidd-Drwg (approved):

PR #46, opened 2018-03-12
Disable unnecessary clang-tidy 6 checks

Felerius:
Clang-tidy 6.0 adds new checks based on the guidelines for googles fuchsia OS. This disables the ones disallowing operator overloads and function calls using default parameters (there might be more unnecessary ones, but these turned up with the current code).


anne-gropler (approved):
Seems reasonable.

PR #47, opened 2018-03-13
Image padding

Blaidd-Drwg:
Depends on #32.

Add a padding argument to the Image(FT_Bitmap) constructor. Padding creates empty space around the original glyph, preventing distance fields from being cut off at the edges.


source/llassetgen/source/Image.cpp
```diff
@@ -72,15 +73,44 @@ namespace llassetgen {
         assert(ft_bitDepth == bitDepth);
 
         stride = size_t(std::ceil(float(getWidth() * bitDepth) / 8));
+        size_t paddingBytes = padding / 8;
+
+        /* because there are (padding % 8) padding bits, each byte from the original bitmap is
```
Lichtso: There is no such thing as left and right, please refer to them by significance (LSBs and MSBs).

source/llassetgen/source/Image.cpp
```diff
@@ -72,15 +73,44 @@ namespace llassetgen {
...
+        // y padding
+        for (size_t y = 0; y < padding; y++) {
             for (size_t x = 0; x < stride; x++) {
-                data.get()[y * stride + x] = 0xFF - ft_bitmap.buffer[y * ft_bitmap.pitch + x];
+                data.get()[y * stride + x] = 0xFF;
+                data.get()[(getHeight() - y - 1) * stride + x] = 0xFF;
+            }
+        }
+
+        for (size_t y = 0; y < ft_bitmap.rows; y++) {
```
Lichtso, line 96: Why don't you use a view to offset the writing, isn't that precisely what they are meant for?

Lichtso:
Replaced by ff972a2be57f321effa02f38bb84e9de0d104adf

PR #48, opened 2018-03-14
Rendering supersampling

anne-gropler:
This pull request implements some features from Issue #36, namely the supersampling as in OpenLL.
That means that the fragment shader code has been reviewed before (you can review it again, if you want; or don't if you are short on time).

In the dropdown menu you can choose the supersampling algorithm. The visual changes are easier to see for small fonts (zoom out using mouse wheel).

This depends on PR #44. (Update: #44 is merged).


source/llassetgen-rendering/WindowQt.h
```diff
@@ -52,6 +52,7 @@ class WindowQt : public QWindow {
     virtual void fontColorGChanged(QString value);
     virtual void fontColorBChanged(QString value);
     virtual void resetTransform3D();
+    virtual void superSamplingChanged(int index);
```
Felerius, line 55: Couldn't this method be defined on the subclass in `main.cpp`? I don't see it used anywhere in the base class.

source/llassetgen-rendering/main.cpp
```diff
@@ -288,6 +298,11 @@ class Window : public WindowQt {
         paint();
     }
 
+    virtual void superSamplingChanged(int index) override {
+        superSampling = (uint)index;
```
Felerius: Should probably be a static_cast

source/llassetgen-rendering/main.cpp
```diff
@@ -307,6 +322,7 @@ class Window : public WindowQt {
     int samplerIndex = 0;
     glm::mat4 transform3D = glm::mat4();
     glm::mat4 projection = glm::perspective(45.f, 1.f, 0.0001f, 100.f);
+    uint superSampling = 0;
```
Felerius: I would either use a C++ unsigned int type if that works, or use `gl::uint` to make clear where `uint` is coming from and why it is used.

Felerius (approved):
I left some small comments, but looks good overall.

source/llassetgen-rendering/WindowQt.h
anne-gropler: That's true, and I already tried that. Seems that Qt couldn't find thos functions as `slots` if they are not defined in the base class. Adding `Q_OBJECT` to the subclass in `main.cpp` results in an error I couldn't fix.
It all comes down to the problem, that some Qt classes want to use their own gl, but I import the gl functions using glbinding. So I have to split the functionality to thos 2 subclasses. Confused? #MeToo.

source/llassetgen-rendering/main.cpp
anne-gropler: uint is from Qt. I use the c++ unsigend int type now, if that helps for readability.

source/llassetgen-rendering/main.cpp
anne-gropler: yep.

PR #49, opened 2018-03-14
switch between rendering distance field and rendering glyphs

anne-gropler:
This pull request implements a feature from Issue #36, namely the ability to toggle between Distance Field and Rendering Glyphs.

You can switch using the button "Distance Field" who renders itself 'activated' when the underlying Distance Field is rendered.

This depends on PR #48 .


source/llassetgen-rendering/WindowQt.h
```diff
@@ -53,6 +53,7 @@ class WindowQt : public QWindow {
     virtual void fontColorBChanged(QString value);
     virtual void resetTransform3D();
     virtual void superSamplingChanged(int index);
+    virtual void toggleDistanceField(bool activated);
```
Felerius, line 56: Defining this in the base class was necessary due to some Qt weirdness, right?

Felerius (approved):
Looks good to me!

source/llassetgen-rendering/WindowQt.h
anne-gropler: Exactly.
Those functions are Qt-Slots and apparently they don't work if they're not defined in base class.

PR #50, opened 2018-03-19
Performance Optimizations & Benchmarks

Lichtso:


source/llassetgen/include/llassetgen/Image.h
```diff
@@ -28,20 +28,32 @@ namespace llassetgen {
 
        public:
         ~Image();
+        Image(const Image&) = delete;
+        Image(Image&& src);
```
Blaidd-Drwg: Shouldn't the assignment operators be deleted / overwritten as well?

source/llassetgen/source/Image.cpp
```diff
@@ -133,6 +123,66 @@ namespace llassetgen {
         }
     }
 
+    template LLASSETGEN_API void Image::centerDownsampling<float>(const Image& src) const;
+    template LLASSETGEN_API void Image::centerDownsampling<uint32_t>(const Image& src) const;
+    template LLASSETGEN_API void Image::centerDownsampling<uint16_t>(const Image& src) const;
+    template LLASSETGEN_API void Image::centerDownsampling<uint8_t>(const Image& src) const;
+    template <typename pixelType>
+    void Image::centerDownsampling(const Image& src) const {
+        assert(src.getWidth()%getWidth() == 0 && src.getHeight()%getHeight() == 0);
```
Blaidd-Drwg, line 165: If we want to downsample single glyphs instead of the entire atlas, this won't always be true, since every glyph's Image can have a different size.

Lichtso: I don't get it. Isn't that the reason why we use views?
Give me an example

Blaidd-Drwg: On the Skype call we discussed two possible solutions:
1) Render a glyph (which might have a size like 257x185), apply distance transform, downsample it (your assert doesn't hold) and put it into a low-res atlas. Pro: Efficient, we only downsample the glyph. Con: harder to implement a downsampling algo.
2) Create a high-res distance field atlas, then downsample it. Pro: power-of-two size, so it's easy to find a scaling factor for which your assert holds. Con: Less efficient, we have to also downsample empty spaces in the atlas. Also, positions might not be integers anymore: the 'A' that was at position 23x30 is suddenly at 11.5x15. I don't know if the FNT format allows for a glyph to start at a fraction of a pixel.

So IIRC we decided on option 1.

Lichtso: We decided to go for option 1.
This means we have to ensure the assert holds when creating the bitmaps (e.g. by adding some padding).

Blaidd-Drwg: Ok, I'll try adding some padding then. This will make the glyph off-center in the Image though, let's hope that's not a problem.

source/llassetgen/include/llassetgen/Image.h
Blaidd-Drwg: `Image& operator=(Image&&)` is missing as well

Blaidd-Drwg:
The second commit breaks the distance transform, merging it into the cli branch produces garbage.
Should be:
atlas

Actual result:
Result

Blaidd-Drwg (approved):

PR #52, opened 2018-03-21
Atlas creation

Blaidd-Drwg:


source/llassetgen/include/llassetgen/Atlas.h
```diff
@@ -0,0 +1,54 @@
...
+#include <llassetgen/Packing.h>
+
+namespace llassetgen {
+    namespace internal {
+        template <class Iter>
+        constexpr int checkImageIteratorType() {
+            using IterTraits = typename std::iterator_traits<Iter>;
+            using IterRefType = typename IterTraits::reference;
+            using IterCategory = typename IterTraits::iterator_category;
+            static_assert(std::is_assignable<Image, IterRefType>::value, "Input elements must be assignable to Image");
```
Felerius: Does this still work with the copy constructor removed? My guess is that this would check if can assign an `Image&` to an `Image`, which would only work with a copy. Otherwise we can always use `std::is_assignable<Image&, IterRefType>`.

source/llassetgen/include/llassetgen/Atlas.h
```diff
@@ -0,0 +1,54 @@
...
+            static_assert(std::is_base_of<std::input_iterator_tag, IterCategory>::value,
+                          "Input iterator must be an InputIterator");
+            return 0;
+        }
+    }
+
+    template <class ImageIter>
+    Image fontAtlas(ImageIter imgBegin, ImageIter imgEnd, Packing packing, uint8_t bitDepth = 1) {
+        internal::checkImageIteratorType<ImageIter>();
+        assert(std::distance(imgBegin, imgEnd) == static_cast<ptrdiff_t>(packing.rects.size()));
```
Felerius: If we want be completely generic we should use `std::iterator_traits<ImageIter>::difference_type` (which is what `std::distance` returns) instead of `ptrdiff_t`. Same goes for `distanceFieldAtlas`.

source/llassetgen/include/llassetgen/Image.h
```diff
@@ -29,18 +29,22 @@ namespace llassetgen {
...
         size_t getBitDepth() const;
+        Vec2<size_t> getSize() const;
         bool isValid(Vec2<size_t> pos) const;
         template <typename pixelType>
         pixelType getPixel(Vec2<size_t> pos) const;
         template <typename pixelType>
         void setPixel(Vec2<size_t> pos, pixelType data) const;
         template <typename pixelType = uint8_t>
         void fillRect(Vec2<size_t> _min, Vec2<size_t> _max, pixelType in = 0) const;
+        void fillPadding(size_t padding);
```
Felerius: Could be private if it is only called from the constructor.

source/tests/llassetgen-tests/Atlas.cpp
```diff
@@ -0,0 +1,37 @@
...
+TEST(AtlasTest, CreateDistanceFieldAtlas) {
+    std::vector<Image> glyphs;
+    glyphs.reserve(atlasTestSizes.size());
+
+    for (const auto& size : atlasTestSizes) {
+        glyphs.emplace_back(size.x, size.y, 1);
+    }
+
+    Packing p = shelfPackAtlas(atlasTestSizes.begin(), atlasTestSizes.end(), false);
+    Image atlas = distanceFieldAtlas<DeadReckoning>(glyphs.begin(), glyphs.end(), p);
```
Felerius: For completeness we should probably factor this test out and call it with both algorithms, but it is probably not that important.

source/llassetgen/include/llassetgen/Atlas.h
Blaidd-Drwg: Good point. Unfortunately `std::is_assignable<Image&, IterRefType>` fails as well, though.

source/llassetgen/include/llassetgen/Atlas.h
Blaidd-Drwg: Might as well, if your type name is shorter than 40 characters you ain't writing no real C++ anyway :grin:

source/llassetgen/include/llassetgen/Atlas.h
Felerius: Maybe replace `Image&` by `const Image&`? Or just use `IterTraits::value_type` instead of `::reference` (can't test this right now unfortunately).

Blaidd-Drwg: Second one worked, thanks :+1:

anne-gropler (approved):

PR #53, opened 2018-03-21
Fix packing bugs and improve tests

Felerius:
Also includes two minor fixes which were supposed to go into the previous commit.

### DIFF
--- a/source/llassetgen/include/llassetgen/packing/internal/MaxRectsPacker.h
+++ b/source/llassetgen/include/llassetgen/packing/internal/MaxRectsPacker.h
@@ -19,7 +19,8 @@ namespace llassetgen {
             bool pack(Rect<PackingSizeType>& rect);
 
            private:
-            LLASSETGEN_NO_EXPORT std::vector<Rect<PackingSizeType>>::iterator findFreeRect(Rect<PackingSizeType>& rect);
+            LLASSETGEN_NO_EXPORT std::vector<Rect<PackingSizeType>>::const_iterator findFreeRect(
+                Rect<PackingSizeType>& rect) const;
             LLASSETGEN_NO_EXPORT void grow();
             LLASSETGEN_NO_EXPORT void cropRects(const Rect<PackingSizeType>& placedRect);
             LLASSETGEN_NO_EXPORT void pruneFreeList();
--- a/source/llassetgen/include/llassetgen/packing/internal/MaxRectsPacker.h
+++ b/source/llassetgen/include/llassetgen/packing/internal/MaxRectsPacker.h
@@ -19,7 +19,7 @@ namespace llassetgen {
             bool pack(Rect<PackingSizeType>& rect);
 
            private:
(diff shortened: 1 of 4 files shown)


source/llassetgen/include/llassetgen/packing/internal/MaxRectsPacker.h
```diff
@@ -19,7 +19,7 @@ namespace llassetgen {
             bool pack(Rect<PackingSizeType>& rect);
 
            private:
-            LLASSETGEN_NO_EXPORT Rect<PackingSizeType>& findFreeRect(Rect<PackingSizeType>& rect);
+            LLASSETGEN_NO_EXPORT std::vector<Rect<PackingSizeType>>::iterator findFreeRect(Rect<PackingSizeType>& rect);
```
Lichtso: Is the value referenced by the iterator actually written somewhere or could it be read-only / const?

source/tests/llassetgen-tests/Packing.cpp
```diff
@@ -130,14 +151,72 @@ class MaxRectsPackingTest : public PackingTest {
...
+    }
 };
 
-TEST_F(MaxRectsPackingTest, TestRejectTooWide) { testRejectTooWide(); }
-TEST_F(MaxRectsPackingTest, TestRejectTooHigh) { testRejectTooHigh(); }
-TEST_F(MaxRectsPackingTest, TestRotateOnly) { testRotateOnly(); }
-TEST_F(MaxRectsPackingTest, TestAcceptAtlasSized) { testAcceptAtlasSized(); }
-TEST_F(MaxRectsPackingTest, TestAcceptMultipleTiny) { testAcceptMultipleTiny(); }
-TEST_F(MaxRectsPackingTest, TestVariableSizePacking) { testVariableSizePacking(); }
+#define ADD_TESTS_FOR_FIXTURE(Fixture)                                              \
```
Lichtso, line 203: Could this be done using templates instead of macros?

source/llassetgen/include/llassetgen/packing/internal/MaxRectsPacker.h
Felerius: Can be const, thanks! Fixed in the latest commit.

source/tests/llassetgen-tests/Packing.cpp
Felerius: I don't think so, unfortunately. The `TEST_F` macro needs to be used in a namespace, so none of the usual recursive inheritance methods for iterating work. Even if it worked, iterating over a list of method names can't be done nicely using templates as far as i know.

One other way to reduce duplication would be to put the test name, test method names in a `*.inc` file and have them expanded using macros. Seems a bit overkill for 8 pairs though.

Blaidd-Drwg (approved):

anne-gropler (approved):
Looks good to me :) and I really like the name changes to camel case :)

Felerius:
The failing AppVeyor build is only a sporadic NuGet error, the other AppVeyor build is working correctly.

PR #55, opened 2018-04-10
Fnt writer

j-tr:


Blaidd-Drwg (commented):
Could you run `make format-llassetgen` to change all the tabs to spaces?

source/llassetgen/source/FntWriter.cpp
```diff
@@ -0,0 +1,215 @@
...
+	void FntWriter::setFontInfo() {
+		// collect font_info 
+		font_info.face = face->family_name + std::string(" ") + face->style_name;
+		font_info.is_bold == false;
+		font_info.is_italic == false;
+		if (face->style_flags == FT_STYLE_FLAG_ITALIC) {
+			font_info.is_italic == true;
+		}
+		if (face->style_flags == FT_STYLE_FLAG_BOLD) {
+			font_info.is_bold == true;
```
Blaidd-Drwg: Some of these `==` should be `=`

source/llassetgen/source/FntWriter.cpp
```diff
@@ -0,0 +1,215 @@
...
+		}
+		int encoding = int(face->charmap->encoding);
+		for (int i = 0; i < 4; ++i) {
+			int mask = 0xFF;
+			font_info.charset += char(encoding & mask);
+			encoding >>= 8;
+		}
+		std::reverse(font_info.charset.begin(), font_info.charset.end());
```
Blaidd-Drwg: You could cast `&encoding` to a `int8_t *` and get the chars from there, then you wouldn't need the mask and the `reverse`

source/llassetgen/source/FntWriter.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+		if (face->charmap->encoding == FT_ENCODING_UNICODE) {
+			font_info.use_unicode = true;
+		}
+		else {
+			font_info.use_unicode = false;
```
Blaidd-Drwg: Can be shortened to `font_info.use_unicode = (face->charmap->encoding == FT_ENCODING_UNICODE)`

source/llassetgen/source/FntWriter.cpp
```diff
@@ -0,0 +1,215 @@
...
+		font_common.is_packed = 0;
+	}
+	
+	void FntWriter::SaveFnt(std::string filepath) {
+		font_common.base = max_y_bearing;
+
+		// check for correct file ending
+		if (filepath.substr(filepath.length() - 4) == ".fnt") {
+			filepath = filepath.substr(0, filepath.length() - 4);
+		}
```
Blaidd-Drwg: Checking for correct user input is something the CLI can do, you can assume you're getting a path ending in ".fnt"

source/llassetgen/source/FntWriter.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+		// cut off path from filename
+		size_t last_slash_pos = filepath.rfind('\\');
+		std::string filename;
+		if (last_slash_pos != std::string::npos) {
+			filename = filepath.substr(last_slash_pos + 1);
+		}
+		else {
+			filename = filepath;
+		}
```
Blaidd-Drwg: `filename` isn't used anywhere

source/llassetgen/source/FntWriter.cpp
```diff
@@ -0,0 +1,215 @@
...
+					kerning_info.first_id = left_gindex;
+					kerning_info.second_id = right_gindex;
+					// kerning is provided in 26.6 fractions of a pixel
+					kerning_info.kerning = float(kerning_vector.x) / 26.6f;
+					kerning_infos.push_back(kerning_info);
+				}
+				right_charcode = FT_Get_Next_Char(face, right_charcode, &right_gindex);
+			}
+
+			left_charcode = FT_Get_Next_Char(face, left_charcode, &left_gindex);
```
Blaidd-Drwg: We don't need kerning info for every single character, just the ones we have in the atlas. I could pass you an iterator as an argument from which you could get the characters

Blaidd-Drwg (changes requested):

source/llassetgen/include/llassetgen/FntWriter.h
```diff
@@ -0,0 +1,87 @@
...
+namespace llassetgen {
+	struct Info {
+		std::string face;
+		int size;
+		bool is_bold;
+		bool is_italic;
+		std::string charset;
+		bool use_unicode;
+		/*
+		float stretch_h;
```
anne-gropler: Friendly reminder: We agreed on camelCase for our variables. (I know that some c++ and freetype variables use underscore, but we already discussed that.)

source/llassetgen/source/FntWriter.cpp
anne-gropler: Even the compiler printed some warnings about that :D

source/llassetgen/source/FntWriter.cpp
anne-gropler: I think it's nice that the lib checks that, too.
Problem I see here is that there is no "else".

source/llassetgen/source/FntWriter.cpp
```diff
@@ -0,0 +1,215 @@
...
+			fnt_file << "kerning "
+				<< "first=" << kerning_info.first_id << " "
+				<< "second=" << kerning_info.second_id << " "
+				<< "amount=" << kerning_info.kerning << std::endl;
+		}
+
+		// close file
+		fnt_file.close();
+	}
+}
```
anne-gropler: Windows user spotted :P :D 
End the file with a new line, please.

source/tests/llassetgen-tests/FntWriter.cpp
```diff
@@ -0,0 +1,50 @@
...
+		position += {1, 1};
+		size += {1, 1};
+		offset += {1.0f, 1.0f};
+	}
+
+	int max_height = 20;
+	writer.setAtlasProperties({ 100, 200 }, max_height);
+
+	writer.SaveFnt(test_destination_path + "fnt.fnt");
+}
```
anne-gropler: new line, see comment above.

source/llassetgen/source/FntWriter.cpp
Blaidd-Drwg: Else isn't needed, the code checks if `filepath` ends in ".fnt" and cuts it off so that line 134 doesn't transform it into "file.fnt.fnt".
IMO the library should not care whether the filename ends with ".fnt" or not, it should save the file where it's told to. The applications can make sure that the library is told a file name that makes sense.

source/llassetgen/include/llassetgen/FntWriter.h
```diff
@@ -0,0 +1,89 @@
...
+		int first_id;
+		int second_id;
+		float kerning;
+	};
+
+	class LLASSETGEN_API FntWriter {
+		public:
+			FntWriter(FT_Face face, std::string face_name, int font_size, float scaling_factor, bool scaled_glyph);
+			void readFont();
+			void setAtlasProperties(Vec2<uint32_t> size, int max_height);
```
Blaidd-Drwg: Should be `PackingSizeType` instead of `uint32_t` to be compatible with the atlas size returned by the packing.

Blaidd-Drwg (approved):

PR #61, opened 2018-04-15
Cli

Blaidd-Drwg:
Depends on #58 and #60


anne-gropler:
Please provide an example cmd line to trigger stuff (PR documentation!)

source/llassetgen-cmd/config/example_atlas_config.ini
```diff
@@ -0,0 +1,12 @@
...
+distfield = parabola
+packing = maxrects
+
+fontname = Arial
+fontsize = 64
+padding = 20
+dynamicrange = -10 20
+
+ascii = true
+glyph = äöüß
```
anne-gropler, line 12: I really appreciate that you prepared such configs! <3

anne-gropler:
Assertion failed: Image.cpp::226
`assert(getWidth() == ft_bitmap.width && getHeight() == ft_bitmap.rows && bitDepth == getFtBitdepth(ft_bitmap));`

Happens when I run the cmd with "`atlas -d parabola --ascii -f Verdana atlas.png`".


When I run in Release mode, no assertion fail (because it's Release Mode...), but also the Image seems ok, apart from that weird lower edge, I am not sure if this is an Issue of the Distance Transform or if this is a consequence of the failed assertion.

source/llassetgen-cmd/source/main.cpp
```diff
@@ -0,0 +1,226 @@
...
+        }
+    }
+    return set;
+}
+
+int parseAtlas(int argc, char** argv) {
+    // Example: llassetgen-cmd atlas -d parabola --ascii -f Verdana atlas.png
+    CLI::App app{atlasHelp};
+
+    // positional arguments
```
anne-gropler: "positional" like in "file path" ? 
I think "positional" might also sound like some x,y-values....

source/llassetgen-cmd/source/main.cpp
```diff
@@ -0,0 +1,226 @@
...
+    unsigned int fontSize = 128;
+    app.add_option("-s, --fontsize", fontSize, fontsizeHelp, true);
+
+    std::string fontName;
+    CLI::Option* fontNameOpt = app.add_option("-f, --fontname", fontName, fontnameHelp);
+
+    std::string fontPath;
+    CLI::Option* fontPathOpt = app.add_option("--fontpath", fontPath, fontpathHelp)->check(CLI::ExistingFile);
+
+    // TODO: app.requires_one
```
anne-gropler: How can this TODO be resolved? Or what you are waiting for to resolve this? (add the answer as comment, so that later developers might understand what you meant with `app.requieres_one`.

source/llassetgen-cmd/source/main.cpp
```diff
@@ -0,0 +1,226 @@
...
+        }
+    } catch (const std::exception& e) {
+        std::cerr << "Error: " << e.what() << std::endl;
+        return 2;
+    }
+
+    return 0;
+}
+
+int parseDistField(int argc, char** argv) {
```
anne-gropler: name: why "parse"? you are not really parsing a distance field here. You are calculating a distance field from a given image.

Ah, I think you mean "parse arguments for dist field creation", right? Same above with "parseAtlas".
Mmh... maybe change name to "parseForDistField" / "parseForAtlas". Or maybe some other grammatical sugar.

Blaidd-Drwg:
Good catch, should be fixed now.

source/llassetgen-cmd/source/main.cpp
Blaidd-Drwg: I use "positional" like the CLI11 lib, I don't know if it's a common term. It means the position of the argument matters, e.g. `cp here there` does something different than `cp there here`. Whereas `cp -r here there` is the same as `cp here -r there` because `-r` is not positional.
A positional is basically any argument that doesn't have `--something` or `-s` in front of it ^^

source/llassetgen-cmd/source/main.cpp
Blaidd-Drwg: parseDistfieldArgs?

source/llassetgen-cmd/source/main.cpp
anne-gropler: ah, that's good to know. Then that's fine as it is.

source/llassetgen-cmd/source/main.cpp
anne-gropler: sounds good!

source/llassetgen-cmd/source/main.cpp
Blaidd-Drwg: Right, I'd forgotten about that. I'm gonna resolve it now, it's quite a severe bug.

anne-gropler (approved):

PR #73, opened 2018-10-26
fix windows deployment and dependencies

anne-gropler:
fix: make qwindows.dll available via CMake variables
fix: newer GLM versions do not initialize matrix with identity
fix: handle Qt Spontaneous Events
fix: small adaptions in shader code and globjects::init, necessary for some drivers

### DIFF
--- a/source/llassetgen-rendering/main.cpp
+++ b/source/llassetgen-rendering/main.cpp
@@ -72,7 +72,7 @@ class Window : public WindowQt {
     virtual bool initializeGL() override {
         globjects::init([this](const char * name) {
             return getProcAddress(name);
-        }, globjects::Texture::BindlessImplementation::Legacy);
+        });
 
         std::cout << std::endl
                   << "OpenGL Version:  " << glbinding::aux::ContextInfo::version() << std::endl
--- a/source/llassetgen-rendering/main.cpp
+++ b/source/llassetgen-rendering/main.cpp
@@ -70,9 +70,7 @@ class Window : public WindowQt {
     virtual ~Window() {}
 
     virtual bool initializeGL() override {
-        globjects::init([this](const char * name) {
-            return getProcAddress(name);
-        });
+        globjects::init(getProcAddress);
 
         std::cout << std::endl
                   << "OpenGL Version:  " << glbinding::aux::ContextInfo::version() << std::endl
@@ -334,7 +332,8 @@ class Window : public WindowQt {
     }
 
     virtual void resetTransform3D() override {
-        transform3D = glm::mat4(1.f);  // set identity        paint();
+        transform3D = glm::mat4(1.f);  // set identity
+        paint();
     }
 
     virtual void triggerNewDT() override {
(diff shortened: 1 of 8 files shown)


source/llassetgen-rendering/main.cpp
```diff
@@ -69,10 +69,10 @@ class Window : public WindowQt {
 
     virtual ~Window() {}
 
-    virtual void initializeGL() override {
+    virtual bool initializeGL() override {
         globjects::init([this](const char * name) {
             return getProcAddress(name);
-        });
+        }, globjects::Texture::BindlessImplementation::Legacy);
```
scheibel:
```suggestion
        });
```

source/llassetgen-rendering/main.cpp
```diff
@@ -195,7 +197,7 @@ class Window : public WindowQt {
         program->setUniform("showDistanceField", showDistanceField);
         program->setUniform("superSampling", superSampling);
         program->setUniform("threshold", dtThreshold);
-
+        vao->bind();
```
scheibel:
```suggestion
```

source/llassetgen-rendering/main.cpp
```diff
@@ -333,7 +335,7 @@ class Window : public WindowQt {
     }
 
     virtual void resetTransform3D() override {
-        transform3D = glm::mat4();  // set identity
+        transform3D = glm::mat4(1.0);  // set identity
```
scheibel:
```suggestion
        transform3D = glm::mat4(1.0f);  // set identity
```

source/llassetgen-rendering/main.cpp
```diff
@@ -408,7 +410,7 @@ class Window : public WindowQt {
     glm::vec4 backgroundColor = glm::vec4(1.f, 1.f, 1.f, 1.f);
     glm::vec4 fontColor = glm::vec4(0.f, 0.f, 0.f, 1.f);
     int samplerIndex = 0;
-    glm::mat4 transform3D = glm::mat4();
+    glm::mat4 transform3D = glm::mat4(1.0);
```
scheibel:
```suggestion
    glm::mat4 transform3D = glm::mat4(1.0f);
```

source/llassetgen-rendering/main.cpp
```diff
@@ -69,10 +69,10 @@ class Window : public WindowQt {
 
     virtual ~Window() {}
 
-    virtual void initializeGL() override {
+    virtual bool initializeGL() override {
         globjects::init([this](const char * name) {
```
scheibel: With current glbinding master, this line will get a problem.
Have a look at https://github.com/cginternals/glbinding/compare/8c18e53faffc4bfda11901c573a5c4bb0fc0b2d3...beac98e46babb819d225dee7bfd9b13774569351 for a workaround.

anne-gropler: What exactly is the problem? This is `globjects::init`, which does not appear in your workaround. Please help me to understand.

scheibel: This language construct dismisses the support for MSVC 2013. As we want to support this compiler further, we were forced to degrade to plain C function pointers for the lookup callback. Unfortunately, the present implementation requires a binding of `this` into the context of the callback that is not possible with plain C function pointers. A workaround is the usage of a global/static variable (as outlined in my reference to glbinding).

anne-gropler: thank you for this explanation, very helpful :)

source/llassetgen-rendering/main.cpp
anne-gropler: @scheibel Just FYI: When I remove this line, I get this error:
```
#fatal: error: 0x502, high severity (API)
Error has been generated. GL error GL_INVALID_OPERATION in (null): (ID: 173538523) Generic error
```
on my machine with those specifications:
```
OpenGL Version:  4.3
OpenGL Vendor:   Intel
OpenGL Renderer: Intel(R) HD Graphics 520
```

However, everything renders fine, so I stay with the removed line.

